PR #4, opened 2017-09-01
Add Userchip component

ahennr:
This PR adds an user chip component to `react-geo`, its tests and an example.


example-templates/basic.html
```diff
@@ -10,20 +10,18 @@
   <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
   <script>/*global hljs*/ hljs.initHighlightingOnLoad();</script>
   <style>
-    #map {
-      height: 256px;
-      margin: 1em;
+    .userchip {
```
dnlkoch: Please rename to more generic class name or move to example code itself.

ahennr: remove style definition in template.

PR #9, opened 2017-09-05
Add ObjectUtil helper class

marcjansen:
This PR adds an `ObjectUtil` helper class from a client project. It is ported with the minimum needed changes, and with additional tests so that coveralls is happy.

Please review.

### DIFF
--- a/src/Util/ObjectUtil.js
+++ b/src/Util/ObjectUtil.js
@@ -22,28 +22,30 @@ class ObjectUtil {
    *     in the provided object.
    *
    * @param {String} queryKey The key to be searched.
-   * @param {Object} [queryObject] The object to be searched on. If not
-   *     provided the global application context (on root-level) will
-   *     be used.
+   * @param {Object} queryObject The object to be searched on
    *
    * @return {*} The target value or `undefined` if the given couldn't be
-   *     found.
+   *     found, or `false` on unexpected input, or an object if a path was
+   *     passed, from which we only could find a part, but not the most specific
+   *     one. TODO Harmonize return values
    */
   static getValue(queryKey, queryObject) {
     var queryMatch;
 
+    if (!isString(queryKey)) {
+      // TODO Replace with new logger after logger implementation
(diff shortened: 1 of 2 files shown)


ahennr:
Basically :+1:

However,  `getValue` method contains some TODOs which maybe can be solved/handled using [`find`](https://lodash.com/docs/4.17.4#find) and [`findLast`](https://lodash.com/docs/4.17.4#findLast) of (Lodash)[https://lodash.com/], what do you think?
/cc @marcjansen @dnlkoch @KaiVolland

src/Util/ObjectUtil.js
```diff
@@ -0,0 +1,110 @@
...
+   * @return {*} The target value or `undefined` if the given couldn't be
+   *     found, or `false` on unexpected input, or an object if a path was
+   *     passed, from which we only could find a part, but not the most specific
+   *     one. TODO Harmonize return values
+   */
+  static getValue(queryKey, queryObject) {
+    var queryMatch;
+
+    if (!isString(queryKey)) {
+      // TODO Replace with new logger after logger implementation
```
KaiVolland: As we got a Logger now please address this util.

src/Util/ObjectUtil.js
```diff
@@ -0,0 +1,110 @@
...
+      // Ext.Logger.error('Missing input parameter queryKey <String>!');
+      // TODO why don't we return undefined?
+      return false;
+    }
+    if (!isObject(queryObject)) {
+      // TODO Replace with new logger after logger implementation
```
KaiVolland: As we got a Logger now please address this util.

marcjansen:
@KaiVolland I have addressed your two comments, would you have another look?

@ahennr 

> However, getValue method contains some TODOs which maybe can be solved/handled using find and findLast of Lodash, what do you think?

OI think before refactoring we should add it as is. If you want me to do it, I can of course go ahead and rewrite this.

PR #11, opened 2017-09-05
Add UndoUtil helper class

marcjansen:
This PR adds an UndoUtil helper class from a client project. It is ported with the minimum needed changes, and without additional tests.

Please review.


ahennr:
LGTM

marcjansen:
Thanks for the review

PR #14, opened 2017-09-06
VisibleComponent HOC

dnlkoch:
This introduces the `VisibleComponent` high order component.

Additionally the transpiled example sources are enhanced for production usage.


src/VisibleComponent/VisibleComponent.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * The HOC factory function.
+ *
+ * Wrapped components will be checked against the activeModules array of
+ * the state: If the wrapped component (identified by it's name) is included
+ * in the state, it will be rendered, if not, it wont.
+ *
+ * @param {Component} WrappedComponent The component to wrap and enhance.
+ * @return {Component} The wrapped component.
```
marcjansen: I think you missed another `@param` here, right?

src/VisibleComponent/VisibleComponent.spec.js
```diff
@@ -0,0 +1,127 @@
...
+import TestUtils from '../Util/TestUtils';
+
+describe('isVisibleComponent', () => {
+  let EnhancedComponent;
+
+  /* eslint-disable require-jsdoc */
+  class MockComponent extends React.Component {
+    render() {
+      return (
+        <div>A mock Component</div>
```
marcjansen: Should we name the file `*.spec.jsx`?

src/VisibleComponent/VisibleComponent.example.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: basic.html
+title: VisibleComponent HOC example
+description: This example shows the usage of the VisibleComponent HOC (High Order Component).
+collection: Examples
+---
+
+This example shows the usage of the VisibleComponent HOC (High Order Component) to
+determine the visibility of a component based on a `activeModules` property. Typically
+this property is managed globally by `react-redux` (or similiar).
```
marcjansen: A little bit more prose would be nice. E.g. we define three compononents, but because of the HOC we only see 2… OK?

src/index.js
```diff
@@ -1,7 +1,9 @@
 import {UserChip} from './UserChip/UserChip.jsx';
 import {Toolbar} from './Toolbar/Toolbar.jsx';
+import {isVisibleComponent} from './VisibleComponent/VisibleComponent.jsx';
```
KaiVolland: No named import needed. After rebasing all the other imports should be unnamed aswell.

KaiVolland:
Good job.

As you introduce new features you should increase the minor version in the package.json.

src/VisibleComponent/VisibleComponent.jsx
dnlkoch: Check.

src/VisibleComponent/VisibleComponent.spec.js
dnlkoch: Check.

src/VisibleComponent/VisibleComponent.example.md
dnlkoch: Check.

src/index.js
dnlkoch: Nope, `isVisibleComponent` is a named export (and should stay as it is as it is a method).

PR #40, opened 2017-09-11
Remove react as normal dependency

KaiVolland:
… and add `npm install react` to travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
 - node
 before_install:
 - npm install -g greenkeeper-lockfile@1
+- npm install react
 before_script:
 - greenkeeper-lockfile-update
 script:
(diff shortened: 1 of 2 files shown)


dnlkoch:
Top! Please merge.

marcjansen:
I think this introduced an error: Locally (with no node modules cached from when we still had react) on Node.js v8.5.0 I can reproduce this error, which is also on travis e.g. https://travis-ci.org/terrestris/react-geo/builds/274727969?utm_source=github_status&utm_medium=notification 

```
ERROR in ./node_modules/antd/lib/avatar/index.js
Module not found: Error: Can't resolve 'react' in '/home/jansen/code/react-geo/node_modules/antd/lib/avatar'
 @ ./node_modules/antd/lib/avatar/index.js 31:13-29
 @ ./src/UserChip/UserChip.jsx
 @ ./src/UserChip/UserChip.spec.jsx
```

This might be because of changes somewhere else, but adding react to the classical dependencies fixes this for me. What do we do here?

PR #62, opened 2017-09-15
Harmonize named imports

### LINKED ISSUE #57: Unify css class-names for components
Every component should have its unique react-geo css-className.
Which could be configured in the upper area of the class definition.

Compare `Legend` class:
TOP:
```javascript
export class Legend extends React.Component {


  /**
   * The className added to this component.
   * @type {String}
   * @private
   */
  className = 'react-geo-legend'
```
in `render()`:
```javascript
const className = this.props.className
  ? `${this.props.className} ${this.className}`
  : this.className;

return (
  <div className={className}>
    <img src={this.state.legendUrl} alt={alt} />
  </div>
```

marcjansen:
This PR suggests to harmonize our imports, so that we have more confidence whether a class we import from 'ol/layer/group'  has to be referred to as `OlLayerGroup` or OlGroupLayer`, for example.

Please review.

Fixes #57

### DIFF
--- a/src/LayerTree/LayerTree.example.jsx
+++ b/src/LayerTree/LayerTree.example.jsx
@@ -3,10 +3,10 @@ import { render } from 'react-dom';
 
 import OlMap from 'ol/map';
 import OlView from 'ol/view';
-import OlGroupLayer from 'ol/layer/group';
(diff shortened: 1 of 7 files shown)


marcjansen:
Thanks for the review.

PR #63, opened 2017-09-15
Introduces unified css-className for current classes

### LINKED ISSUE #57: Unify css class-names for components

KaiVolland:
This introduces a private `className` property and `className` as a react-property for all components.

The pattern of the components className is like: `react-geo-{class}`

A optional className from props will be added to the class to.

closes #57

### DIFF
--- a/src/Button/SimpleButton/SimpleButton.jsx
+++ b/src/Button/SimpleButton/SimpleButton.jsx
@@ -19,10 +19,22 @@ import './SimpleButton.less';
 class SimpleButton extends React.Component {
 
   /**
+   * The className added to this component.
+   * @type {String}
+   * @private
+   */
+  className = 'react-geo-simplebutton'
+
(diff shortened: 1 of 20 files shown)


KaiVolland:
It's defined as a prop in propTypes so it shows up in the doc, but as a global. I'll create an issue for improving the api docs...

Issue #58, opened 2017-09-14
Unify imports


KaiVolland:
We should unify the named imports. Especially for `openlayers`:

1. ol imports by class/folder
Good:
```javascript
import OlLayerTile from 'ol/layer/tile';
```

Bad:
```javascript
import OlTileLayer from 'ol/layer/tile';
```
2. No unneeded fileendings (*.jsx will still need it)
Good:
```javascript
import TestUtils from './TestUtils';
```

Bad:
```javascript
import TestUtils from './TestUtils.js';
```

KaiVolland:
solved with #62

PR #68, opened 2017-09-15
Add LayerTransparencySlider

dnlkoch:
This introduces the `LayerTransparencySlider`.

### DIFF
--- a/src/Slider/LayerTransparencySlider.example.jsx
+++ b/src/Slider/LayerTransparencySlider.example.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { render } from 'react-dom';
+
+import OlMap from 'ol/map';
+import OlView from 'ol/view';
+import OlTileLayer from 'ol/layer/tile';
+import OlOsmSource from 'ol/source/osm';
+import olProj from 'ol/proj';
+
+import LayerTransparencySlider from './LayerTransparencySlider.jsx'; //@react-geo@
+
+//
+// ***************************** SETUP *****************************************
+//
+const layer = new OlTileLayer({
+  name: 'OSM',
+  source: new OlOsmSource()
+});
+const map = new OlMap({
+  layers: [
+    layer
+  ],
+  view: new OlView({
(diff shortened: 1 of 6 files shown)


src/Slider/LayerTransparencySlider.example.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { render } from 'react-dom';
+
+import OlMap from 'ol/map';
+import OlView from 'ol/view';
+import OlTileLayer from 'ol/layer/tile';
+import OlOsmSource from 'ol/source/osm';
```
KaiVolland: Plz use import names as declared in #58 
so `OlTileLayer` would become `OlLayerTile`

dnlkoch: Thanks for the hint!

src/Slider/LayerTransparencySlider.jsx
```diff
@@ -0,0 +1,85 @@
...
+   * Returns the transparency from the provided layer.
+   *
+   * @return {Number} The transparency of the layer.
+   */
+  getLayerTransparency() {
+    // 1 = fully opaque/visible.
+    let opacity = this.props.layer.getOpacity();
+    let transparency = (1 - opacity) * 100;
+    // Remove any digits.
+    transparency = Math.round(transparency);
```
marcjansen: Why? are you sure this is the right thing to do?

src/Slider/LayerTransparencySlider.jsx
```diff
@@ -0,0 +1,85 @@
...
+      ...passThroughProps
+    } = this.props;
+
+    return (
+      <Slider
+        tipFormatter={value => `${value}%`}
+        defaultValue={this.getLayerTransparency()}
+        onChange={(value) => {
+          this.setLayerTransparency(value);
+        }}
```
marcjansen: Should we use `this.setLayerTransparency.bind(this)`, instead of the arrow function?

PR #71, opened 2017-09-15
LayerTree extensions

dnlkoch:
This PR proposes some extensions to the `LayerTree`:
  * Wrapping the `TreeNode` in it's own class `LayerTreeNode`.
    * Currently this component doesn't do anything (except having a custom `className` set), but it may be useful for future development and/or refactoring.
  * Adding the property `nodeTileRenderer`.
    *  `nodeTileRenderer` is an optional function that can be used to pass a custom title to the `TreeNode` (`LayerTreeNode`). As the `TreeNode` is capable of receiving an `ReactNode`, the function can return any renderable element (String, Number, Element etc.) composition. Additionally it receives the layer instance of the current tree node as argument, so any title can do custom layer-specific stuff.
  * Setting `className` and pass filtered props down to `antd` component only.

### DIFF
--- a/src/LayerTreeNode/LayerTreeNode.less
+++ b/src/LayerTreeNode/LayerTreeNode.less
@@ -1,9 +1,3 @@
 .react-geo-layertree-node {
-  display: flex;
-  align-items: center;
 
-  .ant-tree-node-content-wrapper {
-    flex: 1;
-    overflow: hidden;
-  }
 }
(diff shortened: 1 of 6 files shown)


src/LayerTreeNode/LayerTreeNode.less
```diff
@@ -0,0 +1,3 @@
+.react-geo-layertree-node {
+
+}
```
KaiVolland: Plz remove this file if it is not used.

dnlkoch: I agree, but as the `LayerTreeNode` component is a dummy component at the moment, I would leave it here for reference.

Issue #84, opened 2017-09-19
Running npm start fails to compile on master


weskamm:
As the title says, running `npm start` on the current master leads to:

```
ERROR in Map/ScaleCombo/ScaleCombo.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Map/ScaleCombo/ScaleCombo.js:21112,4]

ERROR in Slider/LayerTransparencySlider.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Slider/LayerTransparencySlider.js:21125,4]

ERROR in LayerTree/LayerTree.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][LayerTree/LayerTree.js:20334,4]

ERROR in Legend/Legend.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Legend/Legend.js:18956,4]

ERROR in Panel/Panel/Panel.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Panel/Panel/Panel.js:16712,4]

ERROR in Button/ToggleGroup/ToggleGroup.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Button/ToggleGroup/ToggleGroup.js:16712,4]

ERROR in Toolbar/Toolbar.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Toolbar/Toolbar.js:16712,4]

ERROR in Panel/Titlebar/Titlebar.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Panel/Titlebar/Titlebar.js:16712,4]

ERROR in Button/ToggleButton/ToggleButton.js from UglifyJs
Unexpected token: name (urlParts) [(webpack)-dev-server/client?http:/localhost:8080:24,0][Button/ToggleButton/ToggleButton.js:16712,4]

ERROR in Button/SimpleButton/SimpleButton.js from UglifyJs
[...]

PR #85, opened 2017-09-19
Add link to gh-pages, removed unused command

weskamm:
This removes the `npm start` command as there seems to be no real use for it.
A link to the gh-pages has been added instead.
Should fix https://github.com/terrestris/react-geo/issues/84

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,6 @@ Build everything (outputs will live in `build/`-directory)
 
     npm run build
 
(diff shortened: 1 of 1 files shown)


marcjansen:
One can still call `npm start` and many people will do exactly that without reading the README at all.

That said, I am not opposed to this change but we really should fix the issue raised in #84

weskamm:
I agree #84 still needs to be fixed, the `npm start`command can be added again if the problems are sorted there.
Will merge this now

PR #97, opened 2017-09-21
Update imports in examples to look at index.js

### LINKED ISSUE #88: Fix imports in examples
Currently our examples do the imports like (after the replacement):

```javascript
import SimpleButton from 'react-geo';
```
this is wrong as we need named imports.

We should import the Components from the index.js just like we do it in the tests. We might need to adjust the replacement behaviour for the import string to.

KaiVolland:
Title says it all.

Fixes #88


dnlkoch:
Great!

Please merge.

PR #100, opened 2017-09-22
Fix heap space issues for tests

KaiVolland:
Hopefully this PR will fix the heap space and `no message in 10000 ms` issues we have with the tests.

It adds `max_old_space_size=4096` option to the test scripts.


package.json
```diff
@@ -49,17 +49,18 @@
     "clean:build": "rimraf ./build/*",
     "clean:dist": "rimraf ./dist/*",
     "pretest": "npm run lint",
-    "test": "npm run clean:test && BABEL_ENV=coverage karma start karma.conf.js --single-run",
+    "test": "npm run clean:test && BABEL_ENV=coverage karma start karma.conf.js --single-run -- --max_old_space_size=4096",
+    "test:phantom": "npm run clean:test && BABEL_ENV=coverage karma start karma.conf.js --single-run --browsers PhantomJS -- --max_old_space_size=4096",
```
marcjansen: Tha could probably be simplified to

    npm run test -- --browsers PhantomJS

KaiVolland: No i didn't work because we allready got `-- --max_old_space_size=4096` in `test`.

PR #101, opened 2017-09-22
Basic handling of scale dependant layers

marcjansen:
This makes it so that layers that are shown/hidden because they are scale dependant also have another class in the layer tree node. 

The code contains some TODOs we should tackle later. Generally I would suggest to put some (bigger) part of the LayerTree class into the LayerTreeNode class.

Please review.

### DIFF
--- a/src/LayerTree/LayerTree.jsx
+++ b/src/LayerTree/LayerTree.jsx
@@ -286,26 +286,12 @@ class LayerTree extends React.Component {
     treeNode = <LayerTreeNode
       title={this.getTreeNodeTitle(layer)}
       key={layer.ol_uid}
-      inResolutionRange={this.layerInResolutionRange(layer)}
+      inResolutionRange={MapUtil.layerInResolutionRange(layer, this.props.map)}
     >
       {childNodes}
     </LayerTreeNode>;
 
     return treeNode;
-  }
-
-  /**
-   * [layerInResolutionRange description]
-   * @param  {[type]} layer [description]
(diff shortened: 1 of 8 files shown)


src/LayerTree/LayerTree.jsx
```diff
@@ -158,6 +159,18 @@ class LayerTree extends React.Component {
   }
 
   /**
+   * [registerResolutionChangeHandler description]
+   * @return {[type]} [description]
+   */
```
KaiVolland: Plz fix the jsdocs.

src/LayerTree/LayerTree.jsx
```diff
@@ -280,6 +295,20 @@ class LayerTree extends React.Component {
...
+   * @return {[type]}       [description]
+   */
+  layerInResolutionRange(layer) {
+    const layerMinRes = Math.max(layer.getMinResolution(), -Infinity);
+    const layerMaxRes = Math.min(layer.getMaxResolution(), Infinity);
+    const mapView = this.props.map.getView();
+    const currentRes = mapView.getResolution();
+    const inResolutionRange = currentRes < layerMaxRes && currentRes >= layerMinRes;
+    return inResolutionRange;
+  }
```
KaiVolland: Plz move this to `MapUtil` so it can be reused. And also fix the jsdoc.

marcjansen:
I have done what @KaiVolland suggested and will merge now.

PR #109, opened 2017-09-25
Update peer dependencies (add ol)

KaiVolland:
This makes ol (openlayers) a peer dependency.

Due to several scope-/instanceof-issues with webpack we need to ensure that the react-geo and the project that uses react-geo refer to the same instance of openlayers.

Therefore we make ol a peer dependency.

Additionaly we make the versions of our peer dependencies more lenient as @marcjansen suggested. :top:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
 before_install:
 - npm install -g greenkeeper-lockfile@1
 before_script:
-- npm install --no-save react
+- npm install --no-save ol react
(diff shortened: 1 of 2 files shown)


marcjansen:
OK for me, but our perr dependencies [should be more lenient](https://nodejs.org/uk/blog/npm/peer-dependencies/#using-peer-dependencies).

Please merge once the conflicts are resolved and the versions are less strictly bound to one version.

PR #197, opened 2017-11-09
LayerTree filterFunction

KaiVolland:
This introduces a filterFunction for the LayerTree.

This adds the possibility to exclude layers from the LayerTree by a given filterFunction.


marcjansen:
we need the same for the legend, I guess

KaiVolland:
@marcjansen gonna ad it to the legend now

KaiVolland:
@marcjansen As you create a seperate legend for every layer you don't need a layerfilter inside the legend.

PR #200, opened 2017-11-10
Reworks the ScaleCombo

KaiVolland:
This reworks the ScaleCombo.

Closes #200.


src/Field/ScaleCombo/ScaleCombo.jsx
```diff
@@ -118,15 +179,39 @@ class ScaleCombo extends React.Component {
...
+    this.state.scales.push(roundScale);
   };
 
   /**
+   * Rounds a scalenumber in dependency to its size.
+   *
+   * @param  {Number} scale The exact scale
+   * @return {Number} The roundedScale
+   */
+  roundScale = (scale) => {
```
ahennr: Maybe this can be moved to `MapUtil`?

src/Field/ScaleCombo/ScaleCombo.jsx
```diff
@@ -164,32 +249,11 @@ class ScaleCombo extends React.Component {
...
-    return this.state.scales[this.state.scales.length - 1 - zoom].key;
-  }
-
-  /**
-   * handleOnKeyDown of input:
-   * trigger setScale (passed by props) after ENTER key pressed
-   *
-   * @param {Event} event KeyBoard event
-   */
-  handleOnKeyDown = (event) => {
```
ahennr: I'm not sure whether this was needed for IE11?! Can you check if selection works in IE11 as well if Enter-key is pressed, please?

KaiVolland: Currently none of our examples is working in IE11 so it's hard to debug. I'll merge this and open an Issue regarding IE.

PR #208, opened 2017-11-13
Introduce feature grid

ahennr:
This PR adds a simple feature grid showing the attributes / values of a single features in a table.

### DIFF
--- a/src/Grid/FeatureGrid/FeatureGrid.jsx
+++ b/src/Grid/FeatureGrid/FeatureGrid.jsx
@@ -4,6 +4,7 @@ import {
   Table
 } from 'antd';
 import OlFeature from 'ol/feature';
+import { get } from 'lodash';
 
 import './FeatureGrid.less';
 
@@ -34,6 +35,13 @@ class FeatureGrid extends React.Component {
     attributeNameColumnTitle: PropTypes.string,
 
     /**
+     * Optional value in percent representing the width of the attribute name column
+     * The width of attribute value column wil be calculated depending in this
+     * @type {String}
+     */
+    attributeNameColumnWidthInPercent: PropTypes.number,
+
+    /**
      * Optional title of the attribute value column
      * @type {String}
      */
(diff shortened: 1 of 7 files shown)


src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -0,0 +1,166 @@
...
+  * @param {Array} attributeFilter Array of string values to filter the grid rows
+  */
+  generateFeatureGrid(feature, attributeFilter) {
+    if (!attributeFilter) {
+      attributeFilter = feature.getKeys().filter((attrName) => attrName !== 'geometry');
+    }
+
+    const dataSource = attributeFilter.map((attr) => {
+      const rowObj = {
+        attributeName: attr,
```
dnlkoch: It may be useful to have a mapping function/object to set custom attribute names.

src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -0,0 +1,166 @@
...
+  /**
+  * generateFeatureGrid function
+  * Initialize data store and column definitions of table
+  *
+  * @param {OlFeature} feature feature to display
+  * @param {Array} attributeFilter Array of string values to filter the grid rows
+  */
+  generateFeatureGrid(feature, attributeFilter) {
+    if (!attributeFilter) {
+      attributeFilter = feature.getKeys().filter((attrName) => attrName !== 'geometry');
```
dnlkoch: I think the `geometry` filter should also be applied, if a custom filter is set.

src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -0,0 +1,166 @@
...
+        key: `ATTR_${attr}_fid_${feature.ol_uid}`
+      };
+      return rowObj;
+    });
+
+    const columns = [{
+      title: this.props.attributeNameColumnTitle,
+      dataIndex: 'attributeName',
+      key: 'attributeName',
+      width: '50%'
```
dnlkoch: Should be configurable.

src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -0,0 +1,166 @@
...
+    }, {
+      title: this.props.attributeValueColumnTitle,
+      dataIndex: 'attributeValue',
+      key: 'attributeValue',
+      width: '50%'
```
dnlkoch: Should be configurable.

src/Grid/FeatureGrid/FeatureGrid.jsx
ahennr: `geometry` column will only be shown if user adds it to his column filter array otherwise not.

src/Grid/FeatureGrid/FeatureGrid.jsx
ahennr: Alright, added in eb42715

src/Grid/FeatureGrid/FeatureGrid.jsx
ahennr: Configurable now, see eb42715

ahennr:
THX for review @dnlkoch

PR #214, opened 2017-11-14
Jest

KaiVolland:
This introduces [Jest](https://facebook.github.io/jest/) as the new testing framework.

Additonally to some new config files (see `jest` folder) the main changes to the tests are just syntactically because jest uses as slightliy [different syntax](https://facebook.github.io/jest/docs/en/expect.html) then `expect.js` (CamelCase instead of dot notation): 

This PR also adds `ol` and `react` as `devDependencies`. These are now `devDependencies` aswell as `peerDependencies` what has the advantage that it's no more necessary to install them manually when developing.
Users of `react-geo` will still have to install `ol` and `react` manually and will receive warnings when the didn't.


package.json
```diff
@@ -81,46 +102,41 @@
...
     "less": "2.7.3",
     "less-loader": "4.0.5",
     "metalsmith": "2.3.0",
     "metalsmith-collections": "0.9.0",
     "metalsmith-layouts": "1.8.1",
     "metalsmith-markdown": "0.2.1",
     "minami": "1.2.3",
-    "mocha": "4.0.1",
+    "ol": "4.5.0",
+    "react": "16.1.1",
```
ahennr: OpenLayers and React are already peer dependendies, aren't they?

src/Button/SimpleButton/SimpleButton.spec.jsx
```diff
@@ -9,12 +8,12 @@ import { SimpleButton } from '../../index';
 
 describe('<SimpleButton />', () => {
   it('is defined', () => {
-    expect(SimpleButton).not.to.be(undefined);
+    expect(SimpleButton).not.toBeUndefined();
```
ahennr: Has Jest its own assertion library or is another library than `expect.js` used?

.babelrc
```diff
@@ -5,6 +5,16 @@
     "react"
   ],
   "env": {
+    "test": {
+      "presets":[
+        ["es2015"],
+        "react",
+        "stage-0"
+      ],
```
dnlkoch: I think `presets` aren't needed here. If so, why is `es2015` an array?

package.json
KaiVolland: They where added intentionally -> see PR description.

.babelrc
dnlkoch: Do we really need `stage-0`?

src/Button/SimpleButton/SimpleButton.spec.jsx
KaiVolland: Jest has its own assertion syntax -> see PR description.

jest/__mocks__/fileMock.js
```diff
@@ -0,0 +1,3 @@
+// __mocks__/fileMock.js
```
dnlkoch: Please remove line.

jest/__mocks__/styleMock.js
```diff
@@ -0,0 +1,3 @@
+// __mocks__/styleMock.js
```
dnlkoch: Please remove line.

package.json
ahennr: Oh, I should have read description more carefully. :see_no_evil:  Sorry for the noise

jest/setup.js
```diff
@@ -1,4 +1,5 @@
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import 'whatwg-fetch';
```
dnlkoch: We may want to include `whatwg-fetch` inside the webpack conf instead of importing it here.

package.json
```diff
@@ -61,6 +60,28 @@
...
+    "transformIgnorePatterns": [
+      "node_modules/(?!ol)"
+    ],
+    "setupFiles": [
+      "<rootDir>/jest/__mocks__/shim.js",
+      "<rootDir>/jest/setup.js"
+    ],
+    "collectCoverage": false,
+    "coverageDirectory": "<rootDir>/coverage"
+  },
```
dnlkoch: Please move config to `jest.config.js`.

.travis.yml
```diff
@@ -11,7 +11,6 @@ script:
 after_script:
 - greenkeeper-lockfile-upload
 after_success:
-- npm run-script coveralls
```
dnlkoch: How do we push the coverage to coveralls now?

package.json
```diff
@@ -49,9 +49,8 @@
     "clean:build": "rimraf ./build/*",
     "clean:dist": "rimraf ./dist/*",
     "pretest": "npm run lint",
-    "test": "npm run clean:test && BABEL_ENV=coverage karma start karma.conf.js --single-run -- --max_old_space_size=4096",
-    "test:watch": "karma start karma.conf.js -- --max_old_space_size=4096",
-    "coveralls": "cat coverage/lcov.info | coveralls",
```
dnlkoch: How do we push the coverage to coveralls now?

package.json
```diff
@@ -81,46 +102,41 @@
...
+    "babel-plugin-dynamic-import-node": "1.2.0",
     "babel-plugin-import": "1.6.2",
-    "babel-plugin-istanbul": "4.1.5",
+    "babel-plugin-transform-es2015-modules-commonjs": "6.26.0",
     "babel-plugin-transform-import-extension-jsx-to-js": "0.1.0",
     "babel-polyfill": "6.26.0",
     "babel-preset-es2015": "6.24.1",
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-0": "6.24.1",
-    "coveralls": "3.0.0",
```
dnlkoch: Please keep it, I think we might need it further on.

package.json
```diff
@@ -81,46 +102,41 @@
...
+    "canvas-prebuilt": "1.6.5-prerelease.1",
     "css-loader": "0.28.7",
     "enzyme": "3.1.1",
     "enzyme-adapter-react-16": "1.0.4",
     "eslint": "4.11.0",
     "eslint-plugin-html": "4.0.0",
     "eslint-plugin-markdown": "1.0.0-beta.7",
     "eslint-plugin-react": "7.4.0",
-    "expect.js": "0.3.1",
```
dnlkoch: Yay!

package.json
dnlkoch: I don't like it having them included as dev- and peerDependencies. Let's just create `postinstall` script like this (inside package.json):

```json
{
  "postinstall": "npm i --no-save ol@4 react@16"
}
```

package.json
```diff
@@ -61,6 +60,28 @@
...
+      "\\.(css|less)$": "<rootDir>/jest/__mocks__/styleMock.js"
+    },
+    "transformIgnorePatterns": [
+      "node_modules/(?!ol)"
+    ],
+    "setupFiles": [
+      "<rootDir>/jest/__mocks__/shim.js",
+      "<rootDir>/jest/setup.js"
+    ],
+    "collectCoverage": false,
```
marcjansen: We can probably just override this flag when we are running in a CI envoronment, an then the rest of the coveralls interaction is mostly the same.

package.json
marcjansen: I thought the same, even though I like the public API of expoect a bitt more thant the jest one. But the pros are way more important!

package.json
marcjansen: Whatever @dnlkoch suggests.

jest/setup.js
KaiVolland: I'll open an issue therefore.

KaiVolland:
I think the decrease in the coverage is due to the change of the test environment. Should we merge nevertheless?

marcjansen:
Nice!

PR #223, opened 2017-11-21
LayerTree: Take the layerGroup from the map if not provided

marcjansen:
Titkle says it all. Please review.


dnlkoch:
Thanks, very nice addition!

PR #231, opened 2017-11-23
Adds filter-function to metalsmith

KaiVolland:
This adds a filter function to Metalsmith to avoid copying the spec-files to the examples folder.

### DIFF
--- a/webpack.examples.config.js
+++ b/webpack.examples.config.js
@@ -78,27 +78,4 @@ commonConfig.module = {
   }]
 };
 
-commonConfig.devtool = 'cheap-module-source-map';
-commonConfig.plugins = [
-  ...commonConfig.plugins || [],
-  new webpack.DefinePlugin({
-    'process.env': {
-      NODE_ENV: JSON.stringify('production')
-    }
-  }),
-  new webpack.optimize.UglifyJsPlugin({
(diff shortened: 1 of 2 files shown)


webpack.examples.config.js
```diff
@@ -78,27 +78,4 @@ commonConfig.module = {
...
-      warnings: false,
-      pure_getters: true,
-      unsafe: false,
-      screw_ie8: true
-    },
-    output: {
-      comments: false
-    }
-  })
-];
```
dnlkoch: Why did you remove this?

marcjansen: Yeah, was about to ask the same. without this the FeatureGrid example js clocks in at a smooth 4 MB:

image

KaiVolland: I was just pissed that it took me 5 mins to build the examples.
I didn't check the impact of this.

Even if the FeatureGrid.js only reaches 826KB on my machine without uglifyJS we should not care how long it takes the ci to build the examples, so i readded it.

Maybe we can find a way to set these options environment specific.

marcjansen: We basically have this already. During development we do not want to have the most optimized builds, only when we are building the examples which we want to publish.

The root problem is that the dev setup right now isn't perfect.

PR #241, opened 2017-11-27
ScaleCombo fixes

KaiVolland:
Value of a antd `Select` has to be of type String so we tranform it with `toString()`.


ahennr:
LGTM, plz merge.

PR #243, opened 2017-11-28
Enhances the check if the layergroup prop has changed

KaiVolland:


src/LayerTree/LayerTree.jsx
```diff
@@ -91,9 +91,10 @@ class LayerTree extends React.Component {
     super(props);
 
     this.state = {
-      layerGrop: null,
+      layerGroup: null,
+      layerGroupRevision: null,
       treeNodes: [],
-      checkedKeys: []
+      checkedKeys: [],
```
marcjansen: Hmm. Not sure I like this, even though many do it right now.

marcjansen:
... is this still wip? if so don't bother about my comment

src/LayerTree/LayerTree.jsx
KaiVolland: I think we should not be too strict on this. I see the advantages of the dangling comma in Arrays and objects but i don't use it permanently.

If you disagree and think we should go one way or the other we should add a rule to eslint.

PR #244, opened 2017-11-28
Coordinate reference system combo

ahennr:
This PR introduces a Combobox where uses can choose a coordinate reference system. If no object with predefined CRS definitions (`predefinedCrsDefinitions`) is given in props, the CRS combo can be uses as an `AutoComplete` field querying [epsg.io - API](https://github.com/klokantech/epsg.io#api-for-results) for CRS definitions.

An example shows the usage both, using an array containing predefined values as well as fetching CRS defintions.

In order to handle projections / transformations, `proj4` has been added as dependency. In addition, a`ProjectionUtil` has been added as well.

Plz review.

### DIFF
--- a/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
+++ b/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
@@ -1,14 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {AutoComplete } from 'antd';
+import { AutoComplete } from 'antd';
 const Option = AutoComplete.Option;
-
-import './CoordinateReferenceSystemCombo.less';
 
 import { Logger, UrlUtil } from '../../index.js';
 
 /**
- * Class representating a combo to choose coordinate projection system via a
+ * Class representing a combo to choose coordinate projection system via a
  * dropdown menu and / or autocompletion
  *
  * @class The CoordinateReferenceSystemCombo
@@ -21,7 +19,7 @@ class CoordinateReferenceSystemCombo extends React.Component {
    * @type {String}
    * @private
    */
-  className = 'react-geo-crs-combo'
+  className = 'react-geo-coordinatereferencesystemcombo'
 
   static propTypes = {
     /**
--- a/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
+++ b/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
@@ -93,21 +93,6 @@ class CoordinateReferenceSystemCombo extends React.Component {
   }
 
   /**
-   * This function gets called on success of the EPSG.io fetch.
-   * It simply calls callback function if defined
-   *
-   * @param {Function} callback The callback function to be called
-   * @param {Object} crsDefinitions Array of CRS definitons used in
-   *                 CoordinateReferenceSystemCombo
-   */
-  runViaCallback = (callback, crsDefinitions) => {
-    if (!callback) {
-      return false;
-    }
-    callback(crsDefinitions);
-  }
-
-  /**
    * This function gets called when the EPSG.io fetch returns an error.
    * It logs the error to the console.
    *
@@ -155,7 +140,7 @@ class CoordinateReferenceSystemCombo extends React.Component {
     if (!predefinedCrsDefinitions) {
       this.fetchCrs(value)
         .then(this.transformResults)
-        .then(this.runViaCallback.bind(this, crsDefinitions => this.setState({ crsDefinitions, value })))
+        .then(crsDefinitions => this.setState({crsDefinitions}))
         .catch(this.onFetchError);
     } else {
       this.setState({ value });
--- a/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
+++ b/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
@@ -32,7 +32,7 @@ class CoordinateReferenceSystemCombo extends React.Component {
      * default: http://epsg.io
      * @type {String}
      */
-    epsgIoBaseUrl: PropTypes.string,
+    crsApiUrl: PropTypes.string,
     /**
      * The empty text set if no value is given / provided
      * @type {String}
@@ -56,7 +56,7 @@ class CoordinateReferenceSystemCombo extends React.Component {
 
   static defaultProps = {
     emptyTextPlaceholderText: 'Please select a CRS',
-    epsgIoBaseUrl: 'https://epsg.io/',
+    crsApiUrl: 'https://epsg.io/',
     onSelect: () => {}
   }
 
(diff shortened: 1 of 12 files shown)


src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.less
```diff
@@ -0,0 +1,3 @@
+.react-geo-crs-combo {
+  width: 200px;
```
marcjansen: why do we have a fixed width in the library?

src/Field/NominatimSearch/NominatimSearch.jsx
```diff
@@ -258,16 +258,6 @@ export class NominatimSearch extends React.Component {
...
-      nominatimBaseUrl,
-      format,
-      viewbox,
-      bounded,
-      polygon_geojson,
-      addressdetails,
-      limit,
-      countrycodes,
-      map,
-      onSelect,
```
marcjansen: huh? is this ok here?

src/Util/ProjectionUtil/ProjectionUtil.js
```diff
@@ -0,0 +1,60 @@
...
+export class ProjectionUtil {
+
+  /**
+   * Returns an object of (custom) proj4 definitions.
+   *
+   * @return {Object} The (custom) proj4 definition strings.
+   */
+  static proj4CrsDefinitions() {
+    return {
+      'EPSG:25832': '+proj=utm +zone=32 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'
```
marcjansen: This is not in in the proj database? Or am I misunderstanding stuff here

src/Util/ProjectionUtil/ProjectionUtil.js
```diff
@@ -0,0 +1,60 @@
...
+    };
+  }
+
+  /**
+   * FeatureCollections returned by the GeoServer may be associated with
+   * crs identifiers (e.g. "urn:ogc:def:crs:EPSG::25832") that aren't
+   * supported by proj4 and/or ol3 per default. Add appropriate
```
marcjansen: no version needed, ol is enough

src/Util/ProjectionUtil/ProjectionUtil.js
marcjansen: Is this because we get back the  urn-whatever identifiers? is there some way to look these up as well?

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
...
+     * The className which should be added.
+     * @type {String}
+     */
+    className: PropTypes.string,
+    /**
+     * The API to query for CRS definitions
+     * default: http://epsg.io
+     * @type {String}
+     */
+    epsgIoBaseUrl: PropTypes.string,
```
KaiVolland: The propertyname could be more general like `crsApiUrl`.

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
...
+     * The empty text set if no value is given / provided
+     * @type {String}
+     */
+    emptyTextPlaceholderText: PropTypes.string,
+    /**
+     * An array of predefined crs definitions habving at least value (name of
+     * CRS) and code (e.g. EPSG-code of CRS) property
+     * @type {Array}
+     */
+    predefinedCrsDefinitions: PropTypes.arrayOf(PropTypes.shape({
```
KaiVolland: :+1:

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
...
+   */
+  fetchCrs = (searchVal, callback) => {
+    const { epsgIoBaseUrl } = this.props;
+
+    const queryParameters = {
+      format: 'json',
+      q: searchVal
+    };
+
+    fetch(`${epsgIoBaseUrl}?${UrlUtil.objectToRequestString(queryParameters)}`)
```
KaiVolland: I'd prefer to return the fetch result and work with the result where it is called. This keeps this function cleaner. #cohesion

```javascript
return fetch(`${epsgIoBaseUrl}?${UrlUtil.objectToRequestString(queryParameters)}`)
  .then(response => response.json());
```

See below for the changes to `handleSearch`.

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
...
+    if (!value || value.length === 0) {
+      this.setState({
+        value,
+        crsDefinitions: []
+      });
+      return;
+    }
+
+    if (!predefinedCrsDefinitions) {
+      this.fetchCrs(value, crsDefinitions => this.setState({ crsDefinitions, value })) ;
```
KaiVolland: The changes to `fetchCrs` would lead to further changes here... (see above)

```javascript
this.fetchCrs(value)
  .then(this.transformResults) // to be extracted from onFetchSucess
  .then(crsDefinitions => this.setState({ crsDefinitions, value })) // could be the new onFetchSucces
  .catch(Logger.error(`Error while requesting in CoordinateReferenceSystemCombo: ${error}`)); // could stay in onFetchError
```

src/Field/NominatimSearch/NominatimSearch.jsx
KaiVolland: I think this is mistake… plz revert.

src/Util/ProjectionUtil/ProjectionUtil.js
```diff
@@ -0,0 +1,60 @@
+import proj4 from 'proj4';
+import OlProjection from 'ol/proj';
+
+/**
+ * Helper class for ol/proj4 projection handling.
+ *
+ * @class
+ */
+export class ProjectionUtil {
```
KaiVolland: This util seems to be projectspecific. If we still want it in react-geo `proj4CrsDefinitions` and `proj4CrsMappings` should be removed and instead be passed to the init functions.

src/Field/NominatimSearch/NominatimSearch.jsx
ahennr: I don't see where these variables are used in `render` function?!

KaiVolland: They are not used but you have to extract them from the `passThroughProps` as these props would be all passed to antds `AutoComplete` which does not accept them.

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {AutoComplete } from 'antd';
```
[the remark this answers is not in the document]
ahennr: fixed in 77710baa1d8

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,228 @@
...
+const Option = AutoComplete.Option;
+
+import './CoordinateReferenceSystemCombo.less';
+
+import { Logger, UrlUtil } from '../../index.js';
+
+/**
+ * Class representating a combo to choose coordinate projection system via a
```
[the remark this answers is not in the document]
ahennr: fixed in 77710baa1d8

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.less
ahennr: removed less file in 77710baa1d8. Style can be passed as prop now.

src/Field/NominatimSearch/NominatimSearch.jsx
ahennr: reverted in 0e724b2fbac1e89

src/Util/ProjectionUtil/ProjectionUtil.js
ahennr: :tophat:

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
ahennr: changed in 3cc756ffa0

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
ahennr: changed in 3cc756ffa0

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
ahennr: changed in 3cc756ffa0

src/Util/ProjectionUtil/ProjectionUtil.js
ahennr: Now, crs Mappings and definitions are passed to init functions of util, see 059eaa3

src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx
```diff
@@ -0,0 +1,239 @@
...
+        value,
+        crsDefinitions: []
+      });
+      return;
+    }
+
+    if (!predefinedCrsDefinitions) {
+      this.fetchCrs(value)
+        .then(this.transformResults)
+        .then(this.runViaCallback.bind(this, crsDefinitions => this.setState({ crsDefinitions, value })))
```
KaiVolland: I don't get the sense in the `runViaCallback` method.
```javascript
.then(this.runViaCallback.bind(this, crsDefinitions => this.setState({ crsDefinitions, value })))
```
…should be…

```javascript
.then(crsDefinitions => this.setState({crsDefinitions})) //  value is already set in line 149
```

ahennr: removed in 17775ed

KaiVolland:
Great job. Plz merge.

PR #249, opened 2017-11-29
Makes use of faketimers in circlemenu tests

KaiVolland:
As the circlemenu tests fail randomly due to timing issues this introduces [Jest Faketimers](https://facebook.github.io/jest/docs/en/timer-mocks.html).


ahennr:
LGTM, plz merge.

PR #254, opened 2017-12-01
Measure button component

annarieger:
This PR introduces measure button component, which allows to make measurements on the map.
Currently 3 measure types (`line`, `polygon` and `angle`) are supported. Configuration is possible via property `measureType` e.g.
```
<MeasureButton
   map={map}
   measureType="line"
/>
```

Additionally a new util `MeasureUtil` containing some methods for formatting of measurement results was introduced.

Known issues:

- In multiple drawing mode only label from last drawn feature will be shown currently.

Thanks @KaiVolland for great support :crown: 


src/Button/MeasureButton/MeasureButton.jsx
```diff
@@ -0,0 +1,747 @@
...
+   *
+   * @type {OlFeature}
+   * @private
+   */
+  _feature = null;
+
+  /**
+   * Overlay to show the measurement.
+   *
+   * @type {ol.Overlay}
```
KaiVolland: Plz use either dot or camelcase notation for ol classes. `ol.overlay` vs `OlOverlay`. I'd prefer the second one as we use it like this in react-geo.

src/Button/MeasureButton/MeasureButton.jsx
```diff
@@ -0,0 +1,747 @@
...
+    const {
+      measureType,
+      decimalPlacesInTooltips,
+      map,
+      measureTooltipCssClasses
+    } = this.props;
+
+    if (!isEmpty(this._feature)) {
+      const geom = this._feature.getGeometry();
+      let value = measureType === 'line' ?
```
KaiVolland: You may use `const` here.

src/Button/MeasureButton/MeasureButton.less
```diff
@@ -0,0 +1,60 @@
...
+
+  &.btn-pressed {
+    background-color: darken(@primary-color, 20);
+    border-color: darken(@component-background, 10);
+  }
+}
+
+.react-geo-measure-tooltip {
+  position: relative;
+  background: @measure-dynamic-tooltip-background-color;
```
KaiVolland: :+1:

src/Button/MeasureButton/MeasureButton.spec.jsx
```diff
@@ -0,0 +1,825 @@
...
+      const measureTypeValidValues = [
+        'line',
+        'area',
+        'angle'
+      ];
+
+      expect(measureTypeValidValues).toContain(wrapper.props().measureType);
+
+      let spy = {};
+      spy.console = jest.spyOn(console, 'error').mockImplementation(() => {});
```
KaiVolland: Is there any benefit of creating an object `spy` with property `console` instead of just creating a variable (can be const here btw) named `consoleSpy` or similar?

src/Button/MeasureButton/MeasureButton.jsx
annarieger: I prefer your suggestion too -> adjusted in e9f7173

src/Button/MeasureButton/MeasureButton.jsx
annarieger: Replaced here and somewhere else via 216e403

annarieger:
Many thanks for review and support @KaiVolland!  :pineapple: 

Will merge as soon as travis is happy.

PR #255, opened 2017-12-01
[WIP] Refactors the buttons

KaiVolland:
- Pass evt to Callbacks
- Mark onClick / onToggle with `.isRequired`
- Refactor ToggleButton to include no ToggleGroup specific code
- Enhance docs

### DIFF
--- a/src/Button/MeasureButton/MeasureButton.example.jsx
+++ b/src/Button/MeasureButton/MeasureButton.example.jsx
@@ -48,6 +48,7 @@ render(
       </MeasureButton>
 
       <MeasureButton
+        pressed={true}
         map={map}
         measureType="line"
         showMeasureInfoOnClickedPoints
--- a/src/Button/SimpleButton/SimpleButton.jsx
+++ b/src/Button/SimpleButton/SimpleButton.jsx
@@ -5,9 +5,6 @@ import {
   Tooltip
 } from 'antd';
 import { Icon } from 'react-fa';
-
-import Logger from '../../Util/Logger';
-
 import './SimpleButton.less';
 
 /**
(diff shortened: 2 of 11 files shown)


KaiVolland:
@terrestris/devs review would be nice

src/Button/SimpleButton/SimpleButton.jsx
```diff
@@ -30,25 +27,48 @@ class SimpleButton extends React.Component {
...
+      'bottom',
+      'topLeft',
+      'topRight',
+      'bottomLeft',
+      'bottomRight',
+      'leftTop',
+      'leftBottom',
+      'rightTop',
+      'rightBottom'
+    ])
```
annarieger: Is there some default value for placement? Should it be required?

src/Button/SimpleButton/SimpleButton.spec.jsx
```diff
@@ -40,17 +40,6 @@ describe('<SimpleButton />', () => {
...
-
-    wrapper.find('button').simulate('click');
-
-    expect(logSpy).toHaveProperty('callCount', 1);
-
-    Logger.debug.restore();
-  });
-
   it('calls a given click callback method onClick', () => {
     const onClick = sinon.spy();
```
annarieger: just minor and probably not for now: as we introduced and use `jest` in some other classes, it would be cool to replace `sinon` here as well..

src/Button/ToggleButton/ToggleButton.jsx
```diff
@@ -38,15 +34,63 @@ class ToggleButton extends React.Component {
...
+      'bottom',
+      'topLeft',
+      'topRight',
+      'bottomLeft',
+      'bottomRight',
+      'leftTop',
+      'leftBottom',
+      'rightTop',
+      'rightBottom'
+    ])
```
annarieger: The same as above.. Is there some default value for placement? Should it be required?

src/Button/MeasureButton/MeasureButton.example.jsx
```diff
@@ -48,6 +48,7 @@ render(
       </MeasureButton>
 
       <MeasureButton
```
annarieger: As `MeasureButton` instances are all toggle buttons actually, we could possibly wrap them into a `ToggleGroup` to enhance the example usability.

annarieger:
Nice work, @KaiVolland! I've added some minor notes you maybe like to address.

Please merge at will afterwards :pineapple:

src/Button/SimpleButton/SimpleButton.jsx
KaiVolland: Default is top. See https://ant.design/components/tooltip/

src/Button/ToggleButton/ToggleButton.jsx
KaiVolland: Default is `top`. See https://ant.design/components/tooltip/

src/Button/MeasureButton/MeasureButton.example.jsx
KaiVolland: Sadly the `ToggleGroup` doesn't work with Composed `ToggleButtons`... Maybe the context approach was better...

PR #265, opened 2017-12-05
Refactors the Buttons

KaiVolland:
- Pass evt to Callbacks
- Mark onClick / onToggle with .isRequired
- Enhance docs

### DIFF
--- a/src/Button/ToggleGroup/ToggleGroup.example.jsx
+++ b/src/Button/ToggleGroup/ToggleGroup.example.jsx
@@ -1,9 +1,19 @@
+/* eslint-disable no-console */
 import React from 'react';
 import { render } from 'react-dom';
 import {
   ToggleGroup,
   ToggleButton
 } from '../../index.js';
+
+/**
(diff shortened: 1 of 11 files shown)


src/Button/ToggleGroup/ToggleGroup.example.jsx
```diff
@@ -1,10 +1,20 @@
+/* eslint-disable no-console */
```
ahennr: can `Logger` be used here?

src/Button/ToggleGroup/ToggleGroup.example.jsx
```diff
@@ -1,10 +1,20 @@
...
   ToggleButton
 } from '../../index.js';
 
+/**
+ * A simple handler.
+ * @param {*} pressed
+ * @param {*} evt
+ */
+const onChange = (pressed, evt) => {
+  console.log('ToggleGroup changed', pressed, evt);
```
ahennr: See above comment

PR #289, opened 2017-12-14
Introduce window portal

ahennr:
This PR adds a Window component as a [React Portal](https://reactjs.org/docs/portals.html):
By default, a `<Window>`
*  is always draggable
* is always closable

Herewith, the window can be rendered outside the DOM hierarchy of the parent component (e.g. the application) but it behaves like a normal component in the "main context". In contrast to the former window component included in `<Panel>`, the context of the app is included. This is useful if you want to use a mappified component in the window, for example .

In addition, the layout of the window has been prettyfied :lipstick: by using modal-header class of `antd`.

Plz review.


ahennr:
:top:, thx

KaiVolland:
GJ


Plz dont forget the release. We didn't do a release after the Add-Wms Panel, too.

PR #299, opened 2017-12-20
Enhance development workflow of examples

dnlkoch:
This aims to enhance the (development) workflow of examples by:

  * Automatic rebuild of the example files while hot serving, see new/renamed script `start:examples` so there is no need to rebuild the examples manually anymore.
  * Providing higher build speeds (for development and production as well) due to the usage of shared code parts (chunks).
  * Providing higher build speeds (for development and production as well) due to some configuration cleanups.
  * Build speed on CI comparison:
    * Before ~450s
    * After ~30s


ahennr:
:top: nice!

PR #300, opened 2017-12-20
Greenkeeper/babel jest 22.0.3

dnlkoch:
Follow up of #295.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -975,13 +975,13 @@
       }
     },
     "babel-jest": {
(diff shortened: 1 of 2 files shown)


marcjansen:
Aren't we already at 22.0.3?

dnlkoch:
`jest` was updated to 22.0.3 in #301, but this is about `babel-jest` which is not.

marcjansen:
Thanks for the clarification, easy to get these messed up.

PR #302, opened 2017-12-21
Add some methods to the URLUtil

dnlkoch:
This adds the following methods to the `URLUtil`:
  * getQueryParam()
  * hasQueryParam()
  * createValidGetCapabilitiesRequest()
  * isValid()

### DIFF
--- a/src/Util/UrlUtil/UrlUtil.js
+++ b/src/Util/UrlUtil/UrlUtil.js
@@ -1,5 +1,6 @@
 import URL from 'url-parse';
 import { clone } from 'lodash';
+import isURL from 'validator/lib/isURL';
 
 /**
  * Helper Class for the URL handling.
@@ -53,6 +54,21 @@ export class UrlUtil {
   }
 
   /**
+   * Returns the value of the given query param of the provided URL. If not
+   * found, undefined will be returned.
+   *
+   * @param {String} url The URL to get the query params from.
+   * @param {String} key The key to get the value from.
+   * @return {String} The query param value.
+   */
+  static getQueryParam(url, key) {
+    const queryParamsObj = UrlUtil.getQueryParams(url);
+    const foundKey = Object.keys(queryParamsObj).find(k => k.toLowerCase() === key.toLowerCase());
+
+    return queryParamsObj[foundKey];
+  }
+
+  /**
   * Joins some query parameters (defined by `keys`) of two query objects and
   * returns the joined query parameters.
   *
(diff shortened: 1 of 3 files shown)


src/Util/UrlUtil/UrlUtil.spec.js
```diff
@@ -73,6 +73,22 @@ describe('UrlUtil', () => {
...
+      it('returns the value of the given query param of the provided URL', () => {
+        let got = UrlUtil.getQueryParam('http://borussia.de/bvb', 'bvb');
+        expect(got).toEqual(undefined);
+
+        got = UrlUtil.getQueryParam('http://borussia.de/bvb?bvb=09', 'bvb');
+        expect(got).toEqual('09');
+
+        got = UrlUtil.getQueryParam('http://borussia.de/bvb?bvb=09&shinji=kagawa', 'shinji');
+        expect(got).toEqual('kagawa');
+      });
```
marcjansen: What do we get back from

```
UrlUtil.getQueryParam('http://example.com/foo?humpty=1&humpty=2', 'humpty')
```
?

src/Util/UrlUtil/UrlUtil.spec.js
```diff
@@ -151,6 +211,28 @@ describe('UrlUtil', () => {
...
+      it('validates the given URL', () => {
+        const urls = {
+          'http://ows.terrestris.de': true,
+          'https://ows.terrestris.de': true,
+          'ftp://ows.terrestris.de': true,
+          'http://ows.terrestris.de:8080': true,
+          'http://ows.terrestris.de?param=key': true,
+          'http://ows.terrestris.de?param1=key1&param2=key2': true,
+          'http://10.133.7.9': true,
```
marcjansen: how about testing

```
'http://foo@example.com/foo?humpty=dumpty',
'http://foo:s3cretP4ssw0rd@example.com/foo?humpty=dumpty'
```

do we allow these? I guess the rest of special cases like umlauts and encoded params are handled in the validator library, right?

src/Util/UrlUtil/UrlUtil.spec.js
dnlkoch: Actually it returns `2` as the query string gets parsed by the underlying modules `url-parse`/`querystringify` to `{ humpty: '2' }`.

src/Util/UrlUtil/UrlUtil.js
```diff
@@ -84,6 +100,46 @@ export class UrlUtil {
...
+    const queryParamsObj = UrlUtil.getQueryParams(url);
+
+    return !!Object.keys(queryParamsObj).some(k => k.toLowerCase() === key.toLowerCase());
+  }
+
+  /**
+   * Creates a valid GetCapabilitiesRequest out of the given URL by checking if
+   * SERVICE, REQUEST and VERSION are set.
+   *
+   * @param {String} url The URL to validate. Default is to 'WMS'.
```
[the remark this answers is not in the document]
dnlkoch: Correct, fixed in 2aaa201.

src/Util/UrlUtil/UrlUtil.spec.js
dnlkoch: Yes, both urls are allowed, test cases added in 5dc63fe.

src/Util/UrlUtil/UrlUtil.spec.js
marcjansen: OK. I *think* a return value of `[1, 2]` would be more correct. See e.g. this library: https://github.com/sindresorhus/query-string.

marcjansen:
One other question, shouldn't we use `decodeURIComponent` on the values before returning them? Or is this handled elsewhere?

src/Util/UrlUtil/UrlUtil.spec.js
marcjansen: :+1:

src/Util/UrlUtil/UrlUtil.spec.js
dnlkoch: Yes, I agree. Switched from `querystringify` to `query-string` to get this done. See ca71de7.

dnlkoch:
Thx for review @marcjansen. Will merge this when travis is happy.

PR #304, opened 2017-12-21
The query-string module is a dependency

marcjansen:
Title says it all.

Please review and merge at will.


marcjansen:
Thanks for the review.

PR #315, opened 2018-01-02
Panel - resize prop functions

ahennr:
This PR adds some prop functions (and tests) which are called as soon as the corresponding ones from react-rnd are called. 
 
This can be useful if width and height of a window are needed (e.g. for charting).

Plz review.


ahennr:
thx for review.

PR #318, opened 2018-01-05
Added object update util

hwbllmnn:
Adds a simple object update util function. Also adds canvas as a dev dependency (for me, the tests fail without it).


dnlkoch:
Just wondering, but can't this be achieved by the use of lodash's [set](https://lodash.com/docs/4.17.4#set) or [setWith](https://lodash.com/docs/4.17.4#setWith)?

PR #327, opened 2018-01-09
Enhance parentId check for Window component

annarieger:
This PR adds warning which will be shown, if no valid `parentId` was provided by Window component instantiation and default value `parentId=app` doesnt' match.

If no valid parent was found, the component won't be rendered.

### DIFF
--- a/src/Window/Window.jsx
+++ b/src/Window/Window.jsx
@@ -2,9 +2,9 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import { uniqueId } from 'lodash';
-import { Panel } from  '../index.js';
 
-import { Logger } from '../index';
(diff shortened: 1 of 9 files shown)


src/Window/Window.jsx
```diff
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types';
 import { uniqueId } from 'lodash';
 import { Panel } from  '../index.js';
 
+import { Logger } from '../index';
```
marcjansen: can be put into the above statement... but I always thought we would import from where they are actually first exported. unsure her but also very minor.

annarieger: You're so right! I've fixed imports here and in some other classes in 5ed0cc1

annarieger:
Thanks for review guys! Merge this as travis is happy :pineapple:

PR #339, opened 2018-01-12
Make window compatible to other browsers than chrome

buehner:
The `Window` component did only work in chrome properly. In other browsers (e.g. firefox) the wrapper div of the window has a `style="position: relative;"` element (which seems to come from the parent component [react-rnd](https://github.com/bokuweb/react-rnd)) and the window appeared below the viewport.

This is fixed by adding a `position: inherit !important;` to the `.react-geo-window-portal` CSS class. On top of that, this class will now correctly be set on the wrapper div (but not on the wrapped panel div as before).


ahennr:
:+1:

marcjansen:
Is there any chance to fix this upstream?

PR #355, opened 2018-01-23
Update the map size after update of component

marcjansen:
Please review.


marcjansen:
Thanks for the review.

PR #357, opened 2018-01-24
Enable coverage in Codacy

marcjansen:
This enables coverage in Codacy, but we can only really test this once it is merged, I fear.

I would say this can be merged, and in casse it does not work as exoected I will glady revert this.

What do you think @terrestris/devs?


marcjansen:
I think I have changed my mind and will remove codacy as it does not seem to provuide anythinge we wouldnÄ't have elesewhere, and is also more complicated to set-up.

Issue #206, opened 2017-11-13
InstanceOf / use with Standard ol


KaiVolland:
We use the `PropTypes.instanceOf(<SOMECLASS>)` check in some classes.

I fear that this wont work with the standard ol: `OlMap` != `ol.Map`.
Do we want to support standard ol? If not we have to remove the `basicMap.html` examples...

marcjansen:
What is now an additional module (es modules published as npm package ol) will rather sooner than later become the standard of openlayers.

How likely is it that people build a react app but include a prepared build of openlayers? I guess most will have a dedicated build process which hopefully uses the `ol` package and imports just what is needed.

So I tend to say that I think we do not need to support `ol.Map`, only `OlMap`.

dnlkoch:
Agreed!

KaiVolland:
This is no longer an issue as we removed all apperances of `ol.*` and we only use the package now.

PR #374, opened 2018-01-29
Remove basicMap example template

dnlkoch:
### Description:

This removes the `basicMap.hbs` example template (and any reference to it) as it makes use of the pre-compiled `ol.js`, which we tend to have no support for (see #206).


KaiVolland:
Nice. Plz merge.

PR #379, opened 2018-01-31
Refactors Panel, Titlebar and Window

KaiVolland:
### Description:

This is a breaking change as it removes some props from `Panel`, `Titlebar` and `Window`:

- Removes closable prop from all three classes
  - If you wan't to implement a close button just do it your self and pass it via the `tools` property.
- Removes collapsible prop from Titlebar
  - This prop was not used as the `Panel` passes a CollapseButton if you configure it with `collapsible`
- Removes unused parent property from Titlebar and Panel
- Renames `compress…` props to `collapse…`

It also changes the way the `Panel` component calculates its height and the height of its body.
The old defaults for the props `height` (100) and `width` (100) are replaced with `'auto'` so the Panel will grow with its content, except you specify a specific height and width.

@terrestris/devs plz review (and test)

### DIFF
--- a/src/Panel/Panel/Panel.example.jsx
+++ b/src/Panel/Panel/Panel.example.jsx
@@ -15,43 +15,41 @@ render(
       <Panel
         x={0}
         y={20}
-        title="Collapsible"
-        collapsible={true}
+        title="Children"
       >
-        Content
+        <div style={{padding: '5px'}}>
+          Im a div but i can be any node.
+        </div>
       </Panel>
 
       <Panel
-        x={200}
+        x={250}
         y={20}
+        title="Collapsible"
+        collapsible={true}
+      >
+        <div style={{padding: '5px'}}>
+          Content
+        </div>
+      </Panel>
+
+      <Panel
+        x={420}
+        y={20}
+        width={160}
         collapsible={true}
         title="Tooltip"
         collapseTooltip="Einklappen"
       >
-        Content
-      </Panel>
-
-      <Panel
-        x={400}
-        y={20}
-        title="Children"
-      >
-        <div>Im a div but i can be any node.</div>
+        <div style={{padding: '5px'}}>
+          You can set the tooltip for the collapse icon with the prop `collapseTooltip`.
+        </div>
       </Panel>
 
       <Panel
         x={700}
         y={20}
-        title="Resizable"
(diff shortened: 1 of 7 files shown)


src/Panel/Panel/Panel.jsx
```diff
@@ -306,14 +349,12 @@ export class Panel extends React.Component {
     return (
       <Rnd
         className={rndClassName}
-        ref={(panel) => { this.panel = panel; }}
+        ref={(rnd) => { this.rnd = rnd; }}
         default={{
           x: isNumber(rndOpts.x) ? rndOpts.x : (window.innerWidth / 2) - 160,
           y: isNumber(rndOpts.y) ? rndOpts.y : (window.innerHeight / 2) - 100,
```
dnlkoch: Not related to this PR, but just curious where are these numbers (160 & 100) are from.

src/Panel/Panel/Panel.jsx
```diff
@@ -306,14 +349,12 @@ export class Panel extends React.Component {
     return (
       <Rnd
         className={rndClassName}
-        ref={(panel) => { this.panel = panel; }}
+        ref={(rnd) => { this.rnd = rnd; }}
```
dnlkoch: Parentheses can be omitted.

dnlkoch:
Great one! Only two minor non-blocking comments.

src/Panel/Panel/Panel.jsx
KaiVolland: I don't know.

src/Panel/Panel/Panel.example.jsx
```diff
@@ -12,23 +12,24 @@ render(
...
         x={200}
         y={20}
-        closable={true}
         collapsible={true}
-        title="Tooltips"
-        compressTooltip="Einklappen"
-        closeTooltip="Schließen"
-      />
+        title="Tooltip"
+        collapseTooltip="Einklappen"
```
ahennr: Maybe be could only use english texts in `react-geo`to make texts / tooltips more consistently.

KaiVolland: As this is the explicit example for setting alternate tooltips i used german.
The default value is ofcourse english: `collapse`

PR #390, opened 2018-02-05
Use DOMParser to strip out HTML

marcjansen:
### Description:

I think this is way safer do do it this way, [SO agrees](https://stackoverflow.com/a/47140708). Please review.


KaiVolland:
LGTM. Plz merge.

PR #406, opened 2018-02-15
Greenkeeper/less 3.0.1

KaiVolland:
### Description:

Upgrade less to 3.0.1


dnlkoch:
Unfortunately the current version of `less-loader` isn't compatible with less version 3 at the moment, see [here](https://github.com/webpack-contrib/less-loader/issues/239).

PR #413, opened 2018-02-19
Remove unused i18n dependencies

dnlkoch:
### Description:

`i18next` and `react-i18next` weren't is use and in my opinion `react-geo` should only expose all translatable contents as props so the real translation could take place in the application itself (e.g. by using `translate` provided by `react-i18next`).


KaiVolland:
You're right. I'll merge.

PR #427, opened 2018-02-23
Only apply pressed state to togglebuttons

### LINKED ISSUE #418: SimpleButtons in ToggleGroup
Currently the `SimpleButton` gets a `pressed` prop if inside a `ToggleGroup`.
We need to add a further check in the render method of the `ToggleGroup` to avoid this unwanted behaviour.

KaiVolland:
### Description:

This adds a check to the togglegroup render method to only apply the pressed state to children which are instances of `ToggleButton`.

Fixes #418


annarieger:
LGTM, please merge

PR #433, opened 2018-02-26
Introduces the keyFunction prop to the FeatureGrid

KaiVolland:
### Description:

This introduces the `keyFunction` prop to the `FeatureGrid`. This allows to specify the way the rowkey is generated from the feature.

The default function is set to: `feature => feature.ol_uid`. Which makes this PR an API-breaking change. The previous logic was to just take the row index.

We discovered during project development, that a more precise id would be useful.

### DIFF
--- a/src/Grid/FeatureGrid/FeatureGrid.jsx
+++ b/src/Grid/FeatureGrid/FeatureGrid.jsx
@@ -188,7 +188,16 @@ export class FeatureGrid extends React.Component {
      * The children to render.
      * @type {Element}
      */
-    children: PropTypes.element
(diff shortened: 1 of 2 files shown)


src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -457,7 +467,7 @@ export class FeatureGrid extends React.Component {
   onMapSingleClick = olEvt => {
     const {
       map,
-      features,
+      // features,
```
ahennr: Can this be removed?

PR #447, opened 2018-03-02
Feature grid: rowClassName as function and CSS key fixes

ahennr:
### Description:
* use kebabCase for row keys
  * If not `ol_uid` is used as row key (e.g. by using `keyFunction` prop with `keyFunction={f => f.getId()}` this can lead to errornous row class names that contain irregular characters like `.`. In order to prevent this, keys are [`"kebabCased"`](https://lodash.com/docs/4.17.5#kebabCase).
* enable rowClassName prop being a function as well

### DIFF
--- a/src/Grid/FeatureGrid/FeatureGrid.jsx
+++ b/src/Grid/FeatureGrid/FeatureGrid.jsx
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types';
 import { Table } from 'antd';
 import {
   isEqual,
-  isFunction
+  isFunction,
+  kebabCase
 } from 'lodash';
 import OlStyle from 'ol/style/style';
 import OlStyleFill from 'ol/style/fill';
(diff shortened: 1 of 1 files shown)


src/Grid/FeatureGrid/FeatureGrid.jsx
```diff
@@ -838,9 +843,15 @@ export class FeatureGrid extends React.Component {
       ? `${className} ${this._className}`
       : this._className;
 
-    const finalRowClassName = rowClassName
-      ? `${rowClassName} ${this._rowClassName}`
-      : this._rowClassName;
+    let rowClassNameFn;
+    if (isFunction(rowClassName)) {
+      rowClassNameFn = rowClassName;
```
KaiVolland: I think this ommits `this._rowClassName` which might not be intentional.

ahennr:
IMHO, `kebabCase` only should be applied to generated CSS classes but not on row keys, therefore I've left them as is. If `kebabCase` would be applied (e.g. in `getFeatureFromRowKey`) feature filtering would only work if passed `keyFunction` applies `kebabCase` to it's keys as well, or am I wrong?

src/Grid/FeatureGrid/FeatureGrid.jsx
ahennr: Alright, I've created a function that calls passed function of prop and appends `this._rowClassName`. See 558dc531

ahennr:
THX for your reviews, I'll merge as soon as Travis is happy.

dnlkoch:
Go for it, you're right.

PR #463, opened 2018-03-08
Introduces GeometryUtil

KaiVolland:
### Description:

This introduces the GeometryUtil including these methods:
- splitByLine: Splits an Polygon-Feature or -Geometry into multiple Features or Geometries
:bulb: great job @dnlkoch 
- addBuffer: Adds a Buffer to an Feature or Geometry
- mergeGeometries: Merges mutliple Geometries into on MutliGeometry
- union: Unions multiple Features or Geometries into one Feature or Geometry
- difference: Finds the difference between two polygons by clipping the second polygon from the first.
- intersection: Takes two polygons and finds their intersection.

This introduces [Turf](https://www.npmjs.com/package/@turf/turf) as a dependency

:bulb: **Splitting a geometry with a hole was not yet tested and might be failing**.


src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+ */
+class GeometryUtil {
+
+  /**
+   * Splits a ol.feature with/or ol.geom.Polygon by a ol.feature with/or ol.geom.LineString
+   * into an array of instances of ol.feature with/or ol.geom.Polygon.
+   *
+   * @param {ol.feature | ol.geom.Polygon} polygon The polygon geometry to split.
+   * @param {ol.feature | ol.geom.LineString} lineFeat The line geometry to split the polygon
+   *  geometry with.
```
marcjansen: Will thus also work for multi geometries?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+    // Polygonize the lines.
+    const polygonizedUnionGeom = polygonize(filteredUnionGeom);
+
+    // Return as Array of ol.Features.
+    const features = geoJsonFormat.readFeatures(polygonizedUnionGeom);
+    if (polygon instanceof OlFeature && line instanceof OlFeature) {
+      return features;
+    } else {
+      return features.map(f => f.getGeometry());
+    }
```
marcjansen: we should more clearly document that the return value is dependent on the input values. thinking again I would rather always return one thing, possibly the features. I think this is a better API

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  /**
+   * Adds a buffer to a given geometry.
+   *
+   * @param {ol.geom.Geometry | ol.Feature} geometry The geometry.
+   * @param {Integer} buffer The buffer to add in meters.
+   * @param {String} projection A projection as EPSG code.
```
marcjansen: why would I need that? doc enhancemwnt welcome

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+    } else {
+      return features.map(f => f.getGeometry());
+    }
+  }
+
+  /**
+   * Adds a buffer to a given geometry.
+   *
+   * @param {ol.geom.Geometry | ol.Feature} geometry The geometry.
+   * @param {Integer} buffer The buffer to add in meters.
```
marcjansen: Integer is not a valid type I guess. what happens if a float like number is passed? can we use this to dissolve a feature with a negative value? will buffering consider holes in polygons?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+    const geoJson = geometry instanceof OlFeature
+      ? geoJsonFormat.writeFeatureObject(geometry)
+      : geoJsonFormat.writeGeometryObject(geometry);
+    const buffered = buffer(geoJson, radius, {
+      units: 'meters'
+    });
+    if (geometry instanceof OlFeature) {
+      return geoJsonFormat.readFeature(buffered);
+    } else {
+      return geoJsonFormat.readGeometry(buffered.geometry);
```
marcjansen: I *think* I like one return type better. but this is not blocking this.

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+    geometries.forEach(geom => append(geom));
+    return multiGeom;
+  }
+
+  /**
+   * Takes two or more polygons and returns a combined polygon.
+   *
+   * @param {ol.geom.Geometry[] | ol.Feature[]} polygons An array of ol.Feature
+   *  or ol.geom.Geomerty instances of type polygon.
```
marcjansen: typo. how will multis be handled?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+      const feature = geometry instanceof OlFeature
+        ? geometry
+        : new OlFeature({geometry});
+      if (feature.getGeometry().getType() !== 'Polygon') {
+        invalid = true;
+      }
+      return geoJsonFormat.writeFeatureObject(feature);
+    });
+    if (invalid) {
+      // Logger.warn('Can only create union of polygons.');
```
marcjansen: hmm. why not warn? and why don't we iterate over the single parts of the multipolygon in that case?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+      return undefined;
+    }
+    const unioned = union(...geoJsonsFeatures);
+    const feature = geoJsonFormat.readFeature(unioned);
+    if (polygons[0] instanceof OlFeature) {
+      return feature;
+    } else {
+      return feature.getGeometry();
```
marcjansen: I get it now.  the different return types seem to be everywhere.

src/Util/GeometryUtil/GeometryUtil.spec.js
```diff
@@ -0,0 +1,753 @@
...
+       *   |      |      |
+       *   |      |      |
+       *   |      |      |
+       *   |      |      |
+       *   |      |      |
+       *   +-------------+
+       *          |
+       *          +
+       */
+        it('splits the given convex polygon geometry with a straight line', () => {
```
marcjansen: indentation is of one space I think. Otherwise impressive ascii art 😎

src/Util/GeometryUtil/GeometryUtil.spec.js
```diff
@@ -0,0 +1,753 @@
...
+       *          |
+       *   +-------------+
+       *   |      |      |
+       *   |      |      |
+       *   |      |      |
+       *   |      +---------+
+       *   |             |
+       *   |             |
+       *   +-------------+
+       */
```
marcjansen: indentation

src/Util/GeometryUtil/GeometryUtil.spec.js
```diff
@@ -0,0 +1,753 @@
...
+       * +--------------------------------+
+       *       |    |         |    |
+       *       |    +---------+    |
+       *       |                   |
+       *       |                   |
+       *       |                   |
+       *       |                   |
+       *       +-------------------+
+       *
+       */
```
marcjansen: indentation off. probably some more occurences.

package.json
```diff
@@ -69,6 +69,7 @@
     "react": "~16.0"
   },
   "dependencies": {
+    "@turf/turf": "5.1.6",
```
ahennr: Does `turf` have an integrated validation method, or is it possible to use the one of `jsts` (which is a dependency of `turf` AFAIK)

src/Util/GeometryUtil/GeometryUtil.js
ahennr: Does this work with (Multi-)polygons with holes?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+  static mergeGeometries(geometries) {
+    const geomType = geometries[0].getType();
+    let mixedGeometryTypes = false;
+    geometries.forEach(geometry => {
+      if (geomType !== geometry.getType()) {
+        mixedGeometryTypes = true;
+      }
+    });
+    if (mixedGeometryTypes) {
+      // Logger.warn('Can not merge mixed geometries into one multigeometry.');
```
ahennr: Can this logger be removed / activated?

src/Util/GeometryUtil/GeometryUtil.js
ahennr: Are holes in (Multi)polygons also covered here?

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+   * Finds the difference between two polygons by clipping the second polygon from the first.
+   *
+   * @param {ol.geom.Geometry | ol.Feature} feature An ol.geom.Geoemtry or ol.Feature
+   * @param {ol.geom.Geometry | ol.Feature} feature An ol.geom.Geoemtry or ol.Feature
+   * @param {String} projection A projection as EPSG code.
+   *
+   * @returns {ol.geom.Geometry | ol.Feature} A Feature or Geometry with the area
+   *  of polygon1 excluding the area of polygon2.
+   */
+  static difference(polygon1, polygon2, projection = 'EPSG:3857') {
```
ahennr: It would be possible to pass linestring geometries (e.g.) as well here, wouldn't it? If true, one could be more general here, e.g. ... Find the difference between two geometries ...

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -0,0 +1,292 @@
...
+   * Takes two polygons and finds their intersection.
+   *
+   * @param {ol.geom.Geometry | ol.Feature} feature An ol.geom.Geoemtry or ol.Feature
+   * @param {ol.geom.Geometry | ol.Feature} feature An ol.geom.Geoemtry or ol.Feature
+   * @param {String} projection A projection as EPSG code.
+   *
+   * @returns {ol.geom.Geometry | ol.Feature} A Feature or Geometry with the
+   * shared area of the two polygons or null if the polygons don't intersect.
+   */
+  static intersection(polygon1, polygon2, projection = 'EPSG:3857') {
```
ahennr: See above comment regarding generality.

src/Util/GeometryUtil/GeometryUtil.spec.js
```diff
@@ -0,0 +1,753 @@
...
+            [20, 21],
+            [20, 20]
+          ]]);
+          const intersectionGeometry = GeometryUtil.intersection(poly1, poly2, 'EPSG:4326');
+          expect(intersectionGeometry).toBe(null);
+        });
+      });
+    });
+  });
+});
```
ahennr: :top:

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: Yes it does. I'll add a test.

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: @marcjansen I don't think we do log anywhere in react-geo... We just have `loglevel` as dependency. :see_no_evil: 

@ahennr Yes. I won't add anymore test as these would be more or less just tests agains `turf.js`. But it tested it once in a sandbox.

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: I'd like to leave it as is. We may wan't make use of a logger in the near future and this way we don't forget the loggers here.

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: `turf.js` only supports difference for polygons.

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: `turf.js` only supports difference for intersect.

src/Util/GeometryUtil/GeometryUtil.js
KaiVolland: Docs are enhanced to clearify why we two different return types.

KaiVolland:
I extracted the geometries to a seperate file to increase the readability of the spec file and added some more docs.

package.json
KaiVolland: I don't think turf.js has a validation method. We could a `jsts` as an extra dependency if we want to have validation in react-geo. We can't rely on the existence of `jsts` by using `turf.js` as the dependency might be remove.

PR #468, opened 2018-03-09
Sets filteroption  in ScaleCombo

### LINKED ISSUE #224: ScaleCombo Example - typing in user-scale leads to 1:2
Tested directly on the example named above in FF (55.0.2), same behaviour in Chrome (61.0.3163.79) 

When I type "233455" or even "1 : 233455" into the box (what I can do) and hit enter I end up with a scale of 1:2 - in the box AND for the map.

KaiVolland:
### Description:

This fixes the input filter for the `ScaleCombo` component.

Fixes #224


marcjansen:
👍

Issue #475, opened 2018-03-13
Split Digitize Button


KaiVolland:
We should split the `DigitzeButton` into multiple `buttons`.

At least into `DrawButton` and `ModifyButton`, `CopyButton` and `DeleteButton`.

~~It also contains a prop named `style` which is allready reserverd for css inline-styles. It should be removed. See https://github.com/terrestris/react-geo/blob/master/src/Button/DigitizeButton/DigitizeButton.jsx#L289~~

marcjansen:
Is this still relevant?

PR #480, opened 2018-03-14
Fixes DigitizeButton bugs

KaiVolland:
### Description:

This fixes buggy behaviour of the `DigitizeButton`.
- Most important fix is to not set a style on the feature but use the existing style functions.
- The label is know set as an feature attribute not at the style.

### DIFF
--- a/src/Button/DigitizeButton/DigitizeButton.jsx
+++ b/src/Button/DigitizeButton/DigitizeButton.jsx
@@ -334,8 +334,7 @@ class DigitizeButton extends React.Component {
       digitizeLayer: null,
       interactions: [],
       showLabelPrompt: false,
-      textLabel: '',
-      originalStyle: null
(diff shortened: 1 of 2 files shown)


dnlkoch:
I understand the changes, but could you please explain the buggy behaviour? I think setting the style for each features was intended to allow different styles inside the layer.

Otherwise, LGTM!

KaiVolland:
@dnlkoch 
The `setStyle` on the features where called inside of a styleFunction. :thinking: 
IMHO the styleFunction is the right place to create different styles inside a layer but you should never call `setStyle` but return the appropriate style.

dnlkoch:
Ok, thanks for the clarification!

KaiVolland:
Changing the style per feature should definetly be (re-)implemented later. Maybe we can tackle this together with #475

PR #482, opened 2018-03-14
GeometryUtil - support merge of MultiGeometry

ahennr:
### Description:
This PR adds support of Multi-Geometries (e.g. `MultiPolygon`)  as input  of `mergeGeometries` in `GeometryUtil`.


src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -163,6 +163,32 @@ class GeometryUtil {
...
+          case 'Point':
+            subGeometries.push(...geometry.getPoints());
+            break;
+          default:
+            break;
+        }
+      });
+      geometries = subGeometries;
+      geomType = multiGeomPartType;
+    }
```
dnlkoch: I suggest to move this code to an own method as this might be needed in other util methods inside the `GeometryUtil` as well.

ahennr: Alright, I'll create an issue for this.

PR #485, opened 2018-03-14
Introduces separateGeometries method to GeometryUtil

### LINKED ISSUE #483: GeometryUtil - Create util for splitting MultiLineString | MultiPoint | MultiPolygon

KaiVolland:
### Description:

This introduces separateGeoemtries method to GeometryUtil. To split one or many "MultiGeometries".
The methods accepts a single or an Array of Geometries and/or MultiGeometries and returns an array of single Geometries.

Closes #483

### DIFF
--- a/src/Util/GeometryUtil/GeometryUtil.js
+++ b/src/Util/GeometryUtil/GeometryUtil.js
@@ -26,12 +26,12 @@ import {
 class GeometryUtil {
 
   /**
-   * The Prefix used to detect multi geometries.
+   * The prefix used to detect multi geometries.
    */
-  static MUTLI_GEOM_PREFIX = 'Multi';
+  static MULTI_GEOM_PREFIX = 'Multi';
 
   /**
-   * Splits a ol.feature with/or ol.geom.Polygon by a ol.feature with/or ol.geom.LineString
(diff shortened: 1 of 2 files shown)


src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -211,6 +199,42 @@ class GeometryUtil {
     return multiGeom;
   }
 
+  /**
+   * Splits an array of geometries (and multi geometries) or a single MutliGeom
```
marcjansen: Typo `Mutli`

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -211,6 +199,42 @@ class GeometryUtil {
...
+   * into an array of single geometries.
+   *
+   * @param {ol.geom.Geometry|ol.geom.Geometry[]} geometries An (array of) ol.geom.geometries;
```
marcjansen: An array of ol.geom.geometries or a single geometry, e.g. a multi_geometry.

marcjansen: I think the more appropriate type would be: `ol.geom.SimpleGeometry` since we do not support collections, do we?

PR #509, opened 2018-03-20
Add HOC to enable drag & drop of geojson files

hwbllmnn:
### Description:

This adds a HOC that can be wrapped around a mappified map in order to enable drag & drop of geojson files.

### DIFF
--- a/src/HigherOrderComponent/DropTargetMap/DropTargetMap.example.jsx
+++ b/src/HigherOrderComponent/DropTargetMap/DropTargetMap.example.jsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { render } from 'react-dom';
+import OlSourceOsm from 'ol/source/osm';
+import OlLayerTile from 'ol/layer/tile';
+import OlView from 'ol/view';
+import OlMap from 'ol/map';
+
+import {
+  MapComponent,
+  MapProvider,
+  mappify,
+  onDropAware
+} from '../../index.js';
+
+/**
+ * Create the OlMap (you could do some asynchronus stuff here).
+ *
+ * @return {Promise} Promise that resolves an OlMap.
+ */
+const mapPromise = new Promise((resolve) => {
+  const layer = new OlLayerTile({
+    source: new OlSourceOsm()
+  });
+
+  const map = new OlMap({
+    view: new OlView({
+      center: [
+        135.1691495,
+        34.6565482
+      ],
+      projection: 'EPSG:4326',
+      zoom: 16,
+    }),
+    layers: [layer]
+  });
+
+  resolve(map);
+});
+
+const Map = mappify(onDropAware(MapComponent));
+
+render(
+  <MapProvider map={mapPromise}>
+    <Map style={{
+      height: '512px'
+    }} />
+  </MapProvider>,
+  document.getElementById('exampleContainer')
+);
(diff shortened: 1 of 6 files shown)


src/HigherOrderComponent/DropTargetMap/DropTargetMap.example.jsx
```diff
@@ -0,0 +1,49 @@
...
+      projection: 'EPSG:4326',
+      zoom: 16,
+    }),
+    layers: [layer]
+  });
+
+  resolve(map);
+});
+
+const Map = mappify(onDropAware(MapComponent));
```
marcjansen: Is the order here relevant?

hwbllmnn: I don't think so. The resulting map component will have the onDrop/onDragOver and map properties injected by all wrapped HOCs and the map is only used once the user actually drops a file.

KaiVolland:
Is there any scenario where we want to use this with anything Others then am map? So we could just implement this inside of the map component...

hwbllmnn:
I'm not sure yet what would be the best way how to insert layer metadata editing (layer name, projection etc.), that's why I thought it would best be separated for now. Since you're way more experienced with react, perhaps you've got an idea? Maybe some kind of customizable callback/promise mechanism when wrapping the HOC or something.

KaiVolland:
In my opinion the main purpose of a HOC is it's reusability.

As this HOC only makes sense on a `MapComponent` i'd prefer to put the `onDrop` and `onDragOver` directly into the `MapComponent` and move the read and add methods to an util (read-stuff maybe to new `FileUtil`). The read methods should then not call other methods but return a `Promise` as you allready mentioned.

package.json
```diff
@@ -79,6 +79,7 @@
     "react-dom": "16.2.0",
     "react-fa": "5.0.0",
     "react-rnd": "7.3.1",
+    "shpjs": "^3.4.2",
```
dnlkoch: Please use fixed version.

src/HigherOrderComponent/DropTargetMap/DropTargetMap.jsx
```diff
@@ -0,0 +1,94 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import OlFormatGeoJSON from 'ol/format/geojson';
+import OlLayerVector from 'ol/layer/vector';
+import OlSourceVector from 'ol/source/vector';
+import shp from 'shpjs';
+
+/**
+ * HOC that adds layers to the map if GeoJSON files are dropped on it.
```
dnlkoch: ...or a Shapefile...

KaiVolland:
OK this might be an uncommon but realistic usecase.

But i'd still like to see the methods added to an util class as these could be really usefull in other places.

package.json
hwbllmnn: Nice catch, I hate when that happens via `npm i --save`

hwbllmnn:
Ok, I've factored out the read functions into a new `FileUtil`.

KaiVolland:
Nice. Could you add a test for the Util? :innocent:

hwbllmnn:
I've opened an [issue](#534) for the tests.

PR #523, opened 2018-03-23
Updates/Synchronizes Toolbar styles

KaiVolland:
### Description:

This updates the styles of the `Toolbar`.

The change is small but the impact might be bigger. Any application side fix of the old 'broken' style might cause some ugly behaviour in current applications.
Examples should look the same as they do now.

Not sure how it would be correctly versioned with semver. API will not be broken but it's not a 'safe' update as your application might look different.

### DIFF
--- a/src/Toolbar/Toolbar.less
+++ b/src/Toolbar/Toolbar.less
@@ -3,6 +3,8 @@
   flex-direction: column;
   justify-content: center;
   align-items: center;
+  pointer-events: all;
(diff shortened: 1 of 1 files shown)


marcjansen:
semver wise this fixes an issue people would work around otherwise. i consider this a bugfix. but release notes should mention the change.

annarieger:
Thanks @KaiVolland! I'll merge it now

PR #530, opened 2018-03-25
Use ZoomButoon for both zooming in & out

marcjansen:
## BREAKING CHANGE

* Remove `ZoomInButton` => Use `ZoomButton` with positive delta (the default)
* Remove `ZoomOutButton` => Use `ZoomButton` with negative delta

### Description:

This removes the components `ZoomInButton` and `ZoomOutButton` and introduces
a new component `ZoomButton` which can be configured with a `delta`-prop,
effectively making the aforementioned dedicated components useless.


annarieger:
Very nice fusion! Thanks and please merge :pineapple:

marcjansen:
Thanks for the review.

PR #539, opened 2018-03-29
Added a check for digitizerLayer's existance on the map

amir-ba:
### Description:
This adds a check for the `DigitizeButton` to make sure new interactions are added to the map 

### DIFF
--- a/src/Button/DigitizeButton/DigitizeButton.jsx
+++ b/src/Button/DigitizeButton/DigitizeButton.jsx
@@ -552,12 +552,20 @@ class DigitizeButton extends React.Component {
       drawType,
       map,
       onDrawEnd,
-      onDrawStart
+      onDrawStart,
+      digitizeLayerName
     } = this.props;
 
     let geometryFunction;
     let type = drawType;
+    let digitizeLayer = MapUtil.getLayerByName(map, digitizeLayerName);
 
+    // this checks to see if the digitizeLayer  is in the map 
+    //and if not adds it from the state to the map
+    if (!digitizeLayer) {
+      map.addLayer(this.state.digitizeLayer);
+    } 
+    
     if (drawType === DigitizeButton.RECTANGLE_DRAW_TYPE) {
       geometryFunction = OlInteractionDraw.createBox();
       type = DigitizeButton.CIRCLE_DRAW_TYPE;
--- a/src/Button/DigitizeButton/DigitizeButton.jsx
+++ b/src/Button/DigitizeButton/DigitizeButton.jsx
@@ -560,8 +560,8 @@ class DigitizeButton extends React.Component {
     let type = drawType;
     let digitizeLayer = MapUtil.getLayerByName(map, digitizeLayerName);
 
-    // this checks to see if the digitizeLayer  is in the map 
-    //and if not adds it from the state to the map
+    // this checks to see if the digitizeLayer is in the map 
+    // and if not adds it from the state to the map
     if (!digitizeLayer) {
       map.addLayer(this.state.digitizeLayer);
     } 


src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -552,12 +552,20 @@ class DigitizeButton extends React.Component {
...
     } = this.props;
 
     let geometryFunction;
     let type = drawType;
+    let digitizeLayer = MapUtil.getLayerByName(map, digitizeLayerName);
 
+    // this checks to see if the digitizeLayer  is in the map 
+    //and if not adds it from the state to the map
+    if (!digitizeLayer) {
+      map.addLayer(this.state.digitizeLayer);
```
marcjansen: I don't think we have state in react-geo, @dnlkoch

src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -552,12 +552,20 @@ class DigitizeButton extends React.Component {
...
+      onDrawStart,
+      digitizeLayerName
     } = this.props;
 
     let geometryFunction;
     let type = drawType;
+    let digitizeLayer = MapUtil.getLayerByName(map, digitizeLayerName);
 
+    // this checks to see if the digitizeLayer  is in the map 
+    //and if not adds it from the state to the map
```
marcjansen: whitespace missing

src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -552,12 +552,20 @@ class DigitizeButton extends React.Component {
...
-      onDrawStart
+      onDrawStart,
+      digitizeLayerName
     } = this.props;
 
     let geometryFunction;
     let type = drawType;
+    let digitizeLayer = MapUtil.getLayerByName(map, digitizeLayerName);
 
+    // this checks to see if the digitizeLayer  is in the map
```
marcjansen: too much whitespace

amir-ba: it will take awhile until I get the hang of this whitespace rules :D

src/Button/DigitizeButton/DigitizeButton.jsx
amir-ba: but the component already has a state. I will see @dnlkoch though to get his opinion.

dnlkoch: Everything is fine here, it's the already existing component state, not a global application state.

annarieger: Just wondering why do we need this check here at all.. Actually the availibility of `digitizeLayer` will be ensured while `DigitizeButton` instantiation in `componentDidMount` method by calling `createDigitizeLayer()` function (compare (here)[https://github.com/terrestris/react-geo/blob/master/src/Button/DigitizeButton/DigitizeButton.jsx#L345] and (here)[https://github.com/terrestris/react-geo/blob/master/src/Button/DigitizeButton/DigitizeButton.jsx#L400].
So after this method is called, the `digitizeLayer` should be already written into state and be availiable via `this.state.digitizeLayer`.

It would be nice if you could recheck this again please..

dnlkoch: Ping @amir-ba.

amir-ba: Well this is mainly for the case where the drawing Vector layer is removed from the map for any reason and the component has not changed. This causes the `digitizeLayer` to return null . and nothing is drawn. 
This is at least what I experienced.

marcjansen: that is an error then. is the existance check enough at all? which removed the previousöy created layer?

amir-ba: @marcjansen you are right I later realized the application bug and fixed it but I thought this extra check can't harm anyone and it keeps the consistency of the tool in case these cases happen. Anyhow @annarieger can decide if it is necessary at all to keep it in.

annarieger: Ok, thanks for explanation, I don't mind to leave this check in.

Please fix whitespace issues mentioned from other devs above and merge afterwards, thanks.

PR #547, opened 2018-04-03
Use fixed version instead of repo-url for html-webpack-plugin

KaiVolland:
### Description:

This replaces the repository url with a fixed version (3.1.0) for the `html-webpack-plugin`.
The previous repo does not longer exist which leads to failing builds.


KaiVolland:
549

PR #557, opened 2018-04-04
introduced onKeyDown event listener to handle escape key

annarieger:
### Description:
introduced `onKeyDown` property and event listener to handle pressing of `escape` keyboard key for window or panel component.

Listener will only be activated if the parent panel or window component provides `onEscape` property.

This feature can be useful to close windows or panels by pressing `escape` key instead of clicking close button.

### DIFF
--- a/src/Panel/Panel/Panel.jsx
+++ b/src/Panel/Panel/Panel.jsx
@@ -25,7 +25,15 @@ export class Panel extends React.Component {
    * @type {String}
    * @private
    */
-  className = 'react-geo-panel'
+  className = 'react-geo-panel';
+
+   /**
+    * Printed representation of the pressed escape keyboard key.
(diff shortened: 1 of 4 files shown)


src/Panel/Panel/Panel.jsx
```diff
@@ -351,6 +386,9 @@ export class Panel extends React.Component {
         {titleBar}
         <div
           className="body"
+          tabIndex="0"
+          ref={this.onBodyRef.bind(this)}
```
KaiVolland: No binding needed as you don't use `this` inside the function.

annarieger: Hoppla :eyes: Removed in 74974c0

annarieger:
Thanks, merging..

PR #561, opened 2018-04-06
Fixed tests for CoordinateReferenceSystemCombo and NominatimSearch components

annarieger:
### Description:
Title says it all.
Some tests were failing due to some incorrectly configurated mock objects used inside.

please review @KaiVolland


annarieger:
Thanks, merging

PR #562, opened 2018-04-06
DigitizeButton - rename style prop

ahennr:
### Description:

This PR renames the prop `style` to `drawStyle` since it is reserved for CSS inline-styles. Furthermore, a test has been added checking that passed style property to wrapped `ToggleButton`.

See also last point in #475 .


annarieger:
Nice catch, good to go, thanks :pineapple:

ahennr:
Thx for reviews :top:

PR #563, opened 2018-04-09
Introduces WfsSearch

KaiVolland:
### Description:

This introduces the `WfsSearch`.

### DIFF
--- a/src/Field/WfsSearch/WfsSearch.jsx
+++ b/src/Field/WfsSearch/WfsSearch.jsx
@@ -39,63 +39,68 @@ export class WfsSearch extends React.Component {
      */
     className: PropTypes.string,
     /**
-     * The base URL.
+     * The base URL. Please make sure that the WFS-Server supports CORS.
      * @type {String}
      */
     baseUrl: PropTypes.string.isRequired,
     /**
-     * The Property which should be shown in the inputfield when a selection was
-     * made.
-     */
-    displayField: PropTypes.string,
-    /**
      * The list of attributes that should be searched through.
+     * @type {String[]}
      */
     searchAttributes: PropTypes.arrayOf(PropTypes.string).isRequired,
     /**
      * The namespace URI used for features.
+     * @type {String}
      */
     featureNS: PropTypes.string,
     /**
      * The prefix for the feature namespace.
+     * @type {String}
      */
     featurePrefix: PropTypes.string,
     /**
      * The feature type names. Required.
+     * @type {String[]}
      */
     featureTypes: PropTypes.arrayOf(PropTypes.string).isRequired,
     /**
      * SRS name. No srsName attribute will be set on geometries when this is not
      * provided.
+     * @type {string}
      */
     srsName: PropTypes.string,
     /**
-     * Ther output format of the response.
+     * The output format of the response.
+     * @type {string}
      */
     outputFormat: PropTypes.string,
     /**
      * Maximum number of features to fetch.
+     * @type {number}
      */
     maxFeatures: PropTypes.number,
     /**
      * Geometry name to use in a BBOX filter.
+     * @type {string}
      */
     geometryName: PropTypes.string,
     /**
      * Optional list of property names to serialize.
+     * @type {String[]}
      */
     propertyNames: PropTypes.arrayOf(PropTypes.string),
     /**
      * Filter condition. See http://openlayers.org/en/latest/apidoc/ol.format.filter.html
      * for more information.
+     * @type {object}
      */
     filter: PropTypes.object,
     /**
-     * The ol.map where the map will zoom to.
+     * The ol.map to interact with on selection.
      *
      * @type {Object}
      */
-    map: PropTypes.instanceOf(OlMap).isRequired,
+    map: PropTypes.instanceOf(OlMap),
     /**
      * The minimal amount of characters entered in the input to start a search.
      * @type {Number}
@@ -105,7 +110,9 @@ export class WfsSearch extends React.Component {
      * A render function which gets called with the selected item as it is
      * returned by the server. It must return an `AutoComplete.Option` with
      * `key={feature.id}`.
-     * The default will display the property `name` if existing or the `id` to.
+     * The default will display the property `name` if existing or the `id` to
+     * and requires an `id` field on the feature. A custom function is required
+     * if your features don't have an `id` field.
      *
      * @type {function}
      */
@@ -121,6 +128,7 @@ export class WfsSearch extends React.Component {
     /**
      * Options which are added to the fetch-POST-request. credentials is set to
      * 'same-origin' as default but can be overwritten.
+     * @type {object}
      */
     additionalFetchOptions: PropTypes.object
   }
@@ -137,7 +145,7 @@ export class WfsSearch extends React.Component {
      * @return {AutoComplete.Option} The AutoComplete.Option that will be
      * rendered for each feature.
      */
-    renderOption: (feature) => {
+    renderOption: feature => {
       return (
         <Option key={feature.id}>
           {feature.properties.name ? feature.properties.name : feature.id}
@@ -149,6 +157,7 @@ export class WfsSearch extends React.Component {
      * selected item.
      *
      * @param {object} feature The selected feature as returned by the server.
+     * @param {ol.map} olMap The openlayers map that was passed via prop.
      */
     onSelect: (feature, olMap) => {
       if (feature) {
@@ -200,7 +209,7 @@ export class WfsSearch extends React.Component {
 
     if (value) {
       this.setState({
-        searchTerm: value || ''
+        searchTerm: value
       }, () => {
         if (this.state.searchTerm.length >= this.props.minChars) {
           this.doSearch();
@@ -293,7 +302,7 @@ export class WfsSearch extends React.Component {
   }
 
   /**
-   * This function gets called when the nomintim fetch returns an error.
+   * This function gets called when the WFS GetFeature fetch request returns an error.
    * It logs the error to the console.
    *
    * @param {String} error The errorstring.
@@ -331,7 +340,6 @@ export class WfsSearch extends React.Component {
       additionalFetchOptions,
       baseUrl,
       className,
-      displayField,
       featureNS,
       featurePrefix,
       featureTypes,
--- a/src/Field/WfsSearch/WfsSearch.jsx
+++ b/src/Field/WfsSearch/WfsSearch.jsx
@@ -130,7 +130,12 @@ export class WfsSearch extends React.Component {
      * 'same-origin' as default but can be overwritten.
      * @type {object}
      */
-    additionalFetchOptions: PropTypes.object
+    additionalFetchOptions: PropTypes.object,
+    /**
+     * Options which are passed to the constructor of the ol.format.WFS.
+     * compare: http://openlayers.org/en/latest/apidoc/ol.format.WFS.html
+     */
+    wfsFormatOptions: PropTypes.object
   }
 
   static defaultProps = {
@@ -233,7 +238,8 @@ export class WfsSearch extends React.Component {
       propertyNames,
       srsName,
       baseUrl,
-      additionalFetchOptions
+      additionalFetchOptions,
+      wfsFormatOptions
     } = this.props;
 
     const options = {
@@ -248,7 +254,7 @@ export class WfsSearch extends React.Component {
       filter: this.createFilter()
     };
 
-    const wfsFormat = new OlFormatWFS();
+    const wfsFormat = new OlFormatWFS(wfsFormatOptions);
     const featureRequest = wfsFormat.writeGetFeature(options);
     this.setState({fetching: true});
 
@@ -354,6 +360,7 @@ export class WfsSearch extends React.Component {
       renderOption,
       searchAttributes,
       srsName,
+      wfsFormatOptions,
       ...passThroughProps
     } = this.props;
 
--- a/src/Field/WfsSearch/WfsSearch.jsx
+++ b/src/Field/WfsSearch/WfsSearch.jsx
@@ -1,0 +1,376 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Select,
+  Spin
+} from 'antd';
+const Option = Select.Option;
+
+import Logger from '../../Util/Logger';
+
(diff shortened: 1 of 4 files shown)


src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,376 @@
...
+     * The feature type names. Required.
+     */
+    featureTypes: PropTypes.arrayOf(PropTypes.string).isRequired,
+    /**
+     * SRS name. No srsName attribute will be set on geometries when this is not
+     * provided.
+     */
+    srsName: PropTypes.string,
+    /**
+     * Ther output format of the response.
```
annarieger: Typo `Ther` -> `The`

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,376 @@
...
+     */
+    className: PropTypes.string,
+    /**
+     * The base URL.
+     * @type {String}
+     */
+    baseUrl: PropTypes.string.isRequired,
+    /**
+     * The Property which should be shown in the inputfield when a selection was
+     * made.
```
annarieger: very minor: here and by some further props below `@type` annotation is missing

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,376 @@
...
+    minChars: 3,
+    additionalFetchOptions: {},
+    /**
+     * Create an AutoComplete.Option from the given data.
+     *
+     * @param {Object} feature The feature as returned by the server.
+     * @return {AutoComplete.Option} The AutoComplete.Option that will be
+     * rendered for each feature.
+     */
+    renderOption: (feature) => {
```
annarieger: very minor: `()` around the argument can be ommited

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,376 @@
...
+        <Option key={feature.id}>
+          {feature.properties.name ? feature.properties.name : feature.id}
+        </Option>
+      );
+    },
+    /**
+     * The default onSelect method if no onSelect prop is given. It zooms to the
+     * selected item.
+     *
+     * @param {object} feature The selected feature as returned by the server.
```
annarieger: `olMap` param missing

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,376 @@
...
+    const data = response.features ? response.features : [];
+    this.setState({
+      data,
+      fetching: false
+    });
+  }
+
+  /**
+   * This function gets called when the nomintim fetch returns an error.
+   * It logs the error to the console.
```
annarieger: Please adapt method description

src/Field/WfsSearch/WfsSearch.spec.jsx
```diff
@@ -0,0 +1,169 @@
...
+      //SETUP END
+
+      const wrapper = TestUtil.mountComponent(WfsSearch, {map});
+      const fitSpy = jest.spyOn(map.getView(), 'fit');
+      wrapper.props().onSelect(feature, map);
+      expect(fitSpy).toHaveBeenCalled();
+      setTimeout(() => {
+        expect(map.getView().getCenter()).toEqual([25, 25]);
+        expect(map.getView().getZoom()).toEqual(2);
+      }, 501);
```
annarieger: :sunglasses: :+1:

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+  createFilter() {
+    const {
+      searchTerm
+    } = this.state;
+    const {
+      searchAttributes
+    } = this.props;
+
+    const propertyFilters = searchAttributes.map(attribute =>
+      OlFormatFilter.like(attribute, `*${searchTerm}*`, '*', '.', '!', false));
```
ahennr: What happens here if numerical / boolean attributes are filtered / queried here?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+      featureTypes,
+      geometryName,
+      maxFeatures,
+      outputFormat,
+      propertyNames,
+      srsName,
+      filter: this.createFilter()
+    };
+
+    const wfsFormat = new OlFormatWFS();
```
ahennr: By default `OlFormatWFS` supports WFS version 1.1. Please make the used WFS version configurable (see e.g. [here](https://openlayers.org/en/latest/apidoc/ol.format.WFS.html)).

WFS 2.0.x is not supported, or am I wrong?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+    const featureRequest = wfsFormat.writeGetFeature(options);
+    this.setState({fetching: true});
+
+    fetch(`${baseUrl}`, {
+      method: 'POST',
+      credentials: additionalFetchOptions.credentials
+        ? additionalFetchOptions.credentials
+        : 'same-origin',
+      body: new XMLSerializer().serializeToString(featureRequest),
```
ahennr: I've in mind that there existed some problems using `XMLSerializer` in Firefox.

Did you check this also other browsers like FF, Edge, ..?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+  static propTypes = {
+    /**
+     * A optional className that will be added to the input field.
+     */
+    className: PropTypes.string,
+    /**
+     * The base URL.
+     * @type {String}
+     */
+    baseUrl: PropTypes.string.isRequired,
```
ahennr: Must CORS headers be set by the WFS server?

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: Uff i don't know. I think this depends on the WFS-Server. Either it throws an error or it just won't match... Anything we can do about it?

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: `ol.format.WFS` does not support WFS 2.0.0, is it really necessary to support WFS 1.0.0?

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: No i didn't nice hint. I'll check this.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: Yes, i think so.

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+     * The Property which should be shown in the inputfield when a selection was
+     * made.
+     * @type {String}
+     */
+    displayField: PropTypes.string,
+    /**
+     * The list of attributes that should be searched through.
+     * @type {String[]}
+     */
+    searchAttributes: PropTypes.arrayOf(PropTypes.string).isRequired,
```
marcjansen: Can we get a utility to get these from a `DescribeFeatureType`? Posibbly first fetching these and filtering to those usable?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+     * The base URL.
+     * @type {String}
+     */
+    baseUrl: PropTypes.string.isRequired,
+    /**
+     * The Property which should be shown in the inputfield when a selection was
+     * made.
+     * @type {String}
+     */
+    displayField: PropTypes.string,
```
marcjansen: Can we change this to be a template or a function, so I can show multiple fields and arbitrary text? Also it would be super nice if we could somehow mark the matched textpart in the property.

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+    /**
+     * Maximum number of features to fetch.
+     * @type {number}
+     */
+    maxFeatures: PropTypes.number,
+    /**
+     * Geometry name to use in a BBOX filter.
+     * @type {string}
+     */
+    geometryName: PropTypes.string,
```
marcjansen: A utility to get this from a DescribeFeatureType would be nice

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+     * for more information.
+     * @type {object}
+     */
+    filter: PropTypes.object,
+    /**
+     * The ol.map where the map will zoom to.
+     *
+     * @type {Object}
+     */
+    map: PropTypes.instanceOf(OlMap).isRequired,
```
marcjansen: Why is this required? we can search without a map I guess?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+    minChars: 3,
+    additionalFetchOptions: {},
+    /**
+     * Create an AutoComplete.Option from the given data.
+     *
+     * @param {Object} feature The feature as returned by the server.
+     * @return {AutoComplete.Option} The AutoComplete.Option that will be
+     * rendered for each feature.
+     */
+    renderOption: feature => {
```
marcjansen: If a user can provide this, we should list actual requirements, if any. I am thinking of the id property, e.g.

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+      if (feature) {
+        const olView = olMap.getView();
+        const geoJsonFormat = new OlFormatGeoJSON();
+        const olFeature = geoJsonFormat.readFeature(feature);
+        const geometry = olFeature.getGeometry();
+
+        if (geometry) {
+          olView.fit(geometry, {
+            duration: 500
+          });
```
marcjansen: These options might be a good candidate for configuration, right?

src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -0,0 +1,389 @@
...
+   *                                      is pressed.
+   */
+  onUpdateInput(value) {
+    this.setState({
+      data: []
+    });
+
+    if (value) {
+      this.setState({
+        searchTerm: value || ''
```
marcjansen: `|| ''` seems unneeded

src/Field/WfsSearch/WfsSearch.jsx
marcjansen: no need for 1.0.0 IMO. 2.0.0 would be nice, but is not a must have.

src/Field/WfsSearch/WfsSearch.jsx
marcjansen: for geoserver we can cast e.g. by using a string function like concatenate with an empty string, which will accept any input (according to the docs). this should be configurable defaulting to false probably.

src/Field/WfsSearch/WfsSearch.spec.jsx
```diff
@@ -0,0 +1,169 @@
...
+      });
+      //SETUP END
+
+      const wrapper = TestUtil.mountComponent(WfsSearch, {map});
+      const fitSpy = jest.spyOn(map.getView(), 'fit');
+      wrapper.props().onSelect(feature, map);
+      expect(fitSpy).toHaveBeenCalled();
+      setTimeout(() => {
+        expect(map.getView().getCenter()).toEqual([25, 25]);
+        expect(map.getView().getZoom()).toEqual(2);
```
marcjansen: is this actually tested?

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: Sounds very useful. Can you open an issue for it.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: Sounds very useful. Can you open an issue for it.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: After having a look at the code i found out, that this prop is never used. I'll remove it. All what you expect should be realized in the `renderOption` prop.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: I'll enhance the docs at the property.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: As it's just in the default of the `onSelect` property this doesn't need to be configurable. You can just provide a complete new function if you want a different duration.

src/Field/WfsSearch/WfsSearch.spec.jsx
KaiVolland: `expect.assertions` is missing. I'll add it.

src/Field/WfsSearch/WfsSearch.jsx
KaiVolland: After a quick check, it works as aspected in FireFox and InternetExpolrer.

src/Field/WfsSearch/WfsSearch.jsx
ahennr: Since WFS 1.0.0 will be generated simply by using another GML format in `OlFormatWFS` I would make this configurable

KaiVolland: I :womans_hat: this done in 638c12d

src/Field/WfsSearch/WfsSearch.jsx
marcjansen: Here you are: #567

src/Field/WfsSearch/WfsSearch.jsx
marcjansen: Here you are: #567

PR #575, opened 2018-04-13
Changes to WfsSearch

KaiVolland:
### Description:

This makes adds the `optionLabelProp` to be more flexible. Although it is an antd prop we have to introduce it as a prop because we set a default for it.
The default for the `renderOption` prop has been adapted to the changes.

Breaking change:
Modifies the `searchAttributes` to be receive an object instead of an array. To be more precise doing the WFS-GetFeature request. 

Thx for support @hwbllmnn


hwbllmnn:
Nice, please merge at will!

Issue #432, opened 2018-02-26
Use ag-Grid instead of Ant table


KaiVolland:
As we discovered a lot of limitiations regarding the antd table component during development we should think about changing the underlying grid component to ag-Grid.

KaiVolland:
Missing Features:
- Resizable Columns
- Draggable Columns
- Sort on HeaderClick

dnlkoch:
Yes, for more complex tables we should definitely consider using the `ag-grid` component, but in my opinion we shouldn't replace all `antd` tables blindly without having any need of the more advanced features.

hwbllmnn:
It's implemented in #588 . One thing that's missing: the order of the columns is currently not configurable, since the `columnDefs` prop is an object. In ag-grid the column definitions are an array, so we'll have to create the proper order ourselves when we process the config.

Alternatively we can change the `columnDefs` prop to be an array, breaking backwards compatibility.

hwbllmnn:
I just added support for an `index` property for column definitions to #588 .

dnlkoch:
There are some commented tests in the current implementation. We should fix the issue of the missing `grid` state and reactivate them.

marcjansen:
What is the summarized state here?

dnlkoch:
Feature grid based on `ag-grid` is available and all issues listed here (missing index, disabled tests) are resolved.

PR #588, opened 2018-04-19
Ag featuregrid

hwbllmnn:
### Description:

This adds a feature grid similar to the existing feature grid based on [ag-grid](https://www.ag-grid.com/). Fixes issue #432


KaiVolland:
Could you add one or more screenshots?

dnlkoch:
ag-grid-example

This screenshot shows the example included in this PR while having a selected a hovered feature.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1019 @@
...
+     */
+    onGridIsReady: PropTypes.func
+  };
+
+  /**
+   * The default properties.
+   * @type {Object}
+   */
+  static defaultProps = {
+    theme: 'ag-theme-balham',
```
marcjansen: We should mention that the specific CSS for this theme should be loaded, right?

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1019 @@
...
+
+      if (props[key] instanceof OlGeomGeometry) {
+        return;
+      }
+
+      if (columnDefs[key].index !== undefined) {
+        index = columnDefs[key].index;
+      } else {
+        ++index;
+      }
```
marcjansen: This means that if one column has an index, all of them should have an index, right? We should document that very cleanly

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1019 @@
...
+      const properties = feature.getProperties();
+      const filtered = Object.keys(properties)
+        .filter(key => !(properties[key] instanceof OlGeomGeometry))
+        .reduce((obj, key) => {
+          obj[key] = properties[key];
+          return obj;
+        }, {});
+
+      data.push({
+        key: feature.ol_uid, //this.props.keyFunction(feature),
```
marcjansen: Stray comment?

src/Container/AgFeatureGrid/AgFeatureGrid.spec.jsx
```diff
@@ -0,0 +1,412 @@
...
+  //   wrapper.instance().unhighlightFeatures(features);
+  //
+  //   features.forEach(feature => {
+  //     if (feature.ol_uid === selectedFeatureUid) {
+  //       expect(feature.getStyle()).toEqual(wrapper.prop('selectStyle'));
+  //     } else {
+  //       expect(feature.getStyle()).toBe(null);
+  //     }
+  //   });
+  // });
```
marcjansen: Can we activate the commented out tests now? Here and below.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
dnlkoch: It is already mentioned in the documentation of the corresponding prop, see [here](https://github.com/terrestris/react-geo/pull/588/files#diff-8149a62228de5b566a4a4a816147b181R105).

src/Container/AgFeatureGrid/AgFeatureGrid.spec.jsx
dnlkoch: Tried this, but unfortunately it's not that easy. I added a comment in #432.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
dnlkoch: Currently I'm not even sure, why this has been changed as I think it is still needed. So I would suggest to readd the call to `keyFunction`.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: I've added an appropriate comment on the `columnDefs` prop.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: Done.

package.json
```diff
@@ -67,7 +67,9 @@
   "peerDependencies": {
     "antd": "~3.0",
     "ol": "~4.0",
-    "react": "~16.0"
+    "react": "~16.0",
+    "ag-grid": "17.1.1",
```
KaiVolland: I'd suggest to use `~17` as peerDependencies shouldn't be to strict.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
marcjansen: I know, I was specifically talking about things we need for the default theme.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    }) || [];
+
+    const rowRenderer = grid.api.rowRenderer;
+
+    features.forEach(feature => {
+      const key = this.props.keyFunction(feature);
+
+      let rc;
+      rowRenderer.forEachRowComp((idx, rowComp) => {
+        if (rowComp.getRowNode().data.key === feature.ol_uid) {
```
KaiVolland: This will only work for the default `keyFunction` doesn't it? I think it should be `=== key`

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+        feature.setStyle(selectStyle);
+      } else {
+        feature.setStyle(null);
+      }
+    });
+
+    selectedFeatures.forEach(feature => {
+      let rc;
+      rowRenderer.forEachRowComp((idx, rowComp) => {
+        if (rowComp.getRowNode().data.key === feature.ol_uid) {
```
KaiVolland: This will only work for the default keyFunction doesn't it? I think it should be `=== this.props.keyFunction(feature)`

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    const selectedFeatures = map.getFeaturesAtPixel(olEvt.pixel, {
+      layerFilter: layerCand => layerCand === this._layer
+    }) || [];
+
+    selectedFeatures.forEach(selectedFeature => {
+      const key = this.props.keyFunction(selectedFeature);
+      if (selectedRowKeys.includes(key)) {
+        selectedFeature.setStyle(null);
+
+        const node = this.getRowFromFeatureId(selectedFeature.ol_uid);
```
KaiVolland: See comment on `getRowFromFeatureId` implementation...

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+        if (node) {
+          node.setSelected(false);
+        }
+      } else {
+        selectedFeature.setStyle(selectStyle);
+
+        const node = this.getRowFromFeatureId(selectedFeature.ol_uid);
```
KaiVolland: See comment on `getRowFromFeatureId` implementation...

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    return feature[0];
+  }
+
+  /**
+   * Returns the correspondig rowNode for the given feature id.
+   *
+   * @param {Number} id The feature's unique id to obtain the row from.
+   * @return {Object} The row candidate.
+   */
+  getRowFromFeatureId = id => {
```
KaiVolland: This function has to be renamed/-factored. As the `getRowData` function makes use of the `keyFunction` prop `node.data.key` will no always contain the id.
A more general function `getRowFromFeature` which makes use of the `keyFunction` prop aswell will do the trick.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    } = this.props;
+
+    const row = evt.data;
+    const feature = this.getFeatureFromRowKey(row.key);
+
+    if (isFunction(onRowClick)) {
+      onRowClick(row, feature);
+    }
+
+    this.zoomToFeatures([feature]);
```
KaiVolland: I think the zoom should only happen if no `onRowClick` function is passed as we might allways want to zoom …

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    const {
+      map,
+      highlightStyle
+    } = this.props;
+
+    if (!(map instanceof OlMap)) {
+      return;
+    }
+
+    highlightFeatures.forEach(feature => feature.setStyle(highlightStyle));
```
KaiVolland: Plz make use of a `highlightLayer` to avoid setting the style directly on the feature.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    if (!(map instanceof OlMap)) {
+      return;
+    }
+
+    const selectedRowKeys = this.getSelectedRowKeys();
+
+    unhighlightFeatures.forEach(feature => {
+      const key = this.props.keyFunction(feature);
+      if (selectedRowKeys.includes(key)) {
+        feature.setStyle(selectStyle);
```
KaiVolland: See comment above regarding the `highlightLayer`...

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    const {
+      map,
+      selectStyle
+    } = this.props;
+
+    if (!(map instanceof OlMap)) {
+      return;
+    }
+
+    features.forEach(feature => feature.setStyle(selectStyle));
```
KaiVolland: Plz make use of a `selectLayer`. See comment above regarding the `highlightLayer`...

In the select case it is even more important...

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    const {
+      map,
+      features
+    } = this.props;
+
+    if (!(map instanceof OlMap)) {
+      return;
+    }
+
+    features.forEach(feature => feature.setStyle(null));
```
KaiVolland: See comment above regarding the `selectLayer`...

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+     * @type {String}
+     */
+    className: PropTypes.string,
+
+    /**
+     * The height of the grid.
+     *
+     * @type {Number|String}
+     */
+    height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
KaiVolland: Why do we need this?

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+
+    /**
+     * The width of the grid.
+     *
+     * @type {Number|String}
+     */
+    width: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
KaiVolland: Why do we need this?

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
```diff
@@ -0,0 +1,1021 @@
...
+    }
+
+    return (
+      <div
+        className={finalClassName}
+        style={{
+          height: height,
+          width: width
+        }}
+      >
```
KaiVolland: Why do we need a surrounding div?

src/Container/AgFeatureGrid/AgFeatureGrid.less
```diff
@@ -0,0 +1,5 @@
+@import '../../style/variables.less';
+
+.react-geo-ag-feature-grid {
+
+}
```
KaiVolland: Plz remove this empty file.

src/Container/AgFeatureGrid/AgFeatureGrid.spec.jsx
```diff
@@ -0,0 +1,412 @@
...
+  //   features.forEach(feature => {
+  //     if (selectedRowKeys.includes(feature.ol_uid)) {
+  //       expect(feature.getStyle()).toEqual(wrapper.prop('selectStyle'));
+  //     } else {
+  //       expect(feature.getStyle()).toBe(null);
+  //     }
+  //   });
+  //
+  //   expect(wrapper.state('selectedRowKeys')).toEqual(selectedRowKeys);
+  // });
```
KaiVolland: Plz remove outcommented tests if possible.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
dnlkoch: The CSS for the default theme is imported [here](https://github.com/terrestris/react-geo/pull/588/files#diff-8149a62228de5b566a4a4a816147b181R25).

package.json
hwbllmnn: Done.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: Nice catch! Fixed in 85cacd7 .

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: See above

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: See above

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: See above

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: I've renamed the function, actually it was only used with the key as the argument.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: True, I moved it into the else block.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: True. That code has been taken over from the original `FeatureGrid`. I've opened #594 to refactor/fix this.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: See above.

src/Container/AgFeatureGrid/AgFeatureGrid.jsx
hwbllmnn: This fits the default ag-grid behaviour to match its parent div's height and width. I'm sure there are other/better ways to do this, if you have a specific approach in mind.

KaiVolland: Not yet. Just go ahead with the `width` and `height`...

PR #590, opened 2018-04-20
Normalize button layout/styles

annarieger:
### Description:
* Synchronized styles of `SimpleButton` und `ToggleButton` components
* Removed unnecessary CSS rules regarding button elements styling  for `MeasureButton` and `DigitizeButton` as these will be inherited from parent components (e.g. `ToggleButton`) per default.

please review @terrestris/devs


marcjansen:
nice clean-up

PR #602, opened 2018-04-25
Refactor AgFeatureGrid

KaiVolland:
### Description:

- This moves the AgFeatureGrid to the grid folder
- It also moves `ag-grid` and `ag-grid-react` from `peerDependencies` too `dependencies`


package.json
```diff
@@ -67,12 +67,12 @@
...
     "ol": "~4.0",
-    "react": "~16.0",
-    "ag-grid": "~17",
-    "ag-grid-react": "~17"
+    "react": "~16.0"
   },
   "dependencies": {
     "@turf/turf": "5.1.6",
+    "ag-grid": "17.1.1",
+    "ag-grid-react": "17.1.0",
```
dnlkoch: `17.1.1` isn't available here?

KaiVolland: nope not yet

PR #605, opened 2018-04-26
Fixes buggy behaviour when toggleButton is pressed

amir-ba:
### Description:
This fixes a buggy behaviour of the `ToggleButton`.
The bug appears when the component's pressed property has not changed but the pressed state has changed and the `componentWillReceiveProps` function uses the initial props to trigger the button.
With this fix we will check if the props have changed at all before we check the new props against the current state.

Plz review.

### DIFF
--- a/src/Button/ToggleButton/ToggleButton.jsx
+++ b/src/Button/ToggleButton/ToggleButton.jsx
@@ -93,7 +93,8 @@ class ToggleButton extends React.Component {
    * @param {Object} nextProps The received properties.
    */
   componentWillReceiveProps(nextProps) {
-    if (this.state.pressed !== nextProps.pressed) {
+    // Checks if the pressed property has changed before checking its match with the current state 
+    if (this.props.pressed !== nextProps.pressed && this.state.pressed !== nextProps.pressed) {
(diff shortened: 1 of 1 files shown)


amir-ba:
@dnlkoch I agree. The code seems to be patchy. It dose need a rewrite.

src/Button/ToggleButton/ToggleButton.jsx
```diff
@@ -93,7 +93,8 @@ class ToggleButton extends React.Component {
    * @param {Object} nextProps The received properties.
    */
   componentWillReceiveProps(nextProps) {
-    if (this.state.pressed !== nextProps.pressed) {
+    // Checks if the pressed property has changed before checking its match with the current state 
+    if (this.props.pressed !== nextProps.pressed && this.state.pressed !== nextProps.pressed) {
```
KaiVolland: The state should always represent the correct status so the previous check should be enough if the state is always set correctly.

KaiVolland:
IMHO we can close this PR and make a complete refactoring in the near future: #613

amir-ba:
Closing the PR and opting for a refactoring.

PR #614, opened 2018-05-02
Use `jest.DoneCallback` in test of WfsSearch

ahennr:
### Description:

If this is not included local tests for `WfsSearch` run into an error
```
Ran 100000 timers, and there are still more! Assuming we've hit an infinite
recursion and bailing out...
```
 Using this fix, Jest will wait until the `done` callback is called before finishing the test.


annarieger:
LGTM! Nice catch :pineapple:

ahennr:
:top:

Issue #621, opened 2018-05-03
WfsSearch - buggy response handling


KaiVolland:
Implement mechanism to abort pending requests or ensure that the repsonse of the latest request will be shown.

e.g.
When searching for alabama:

Typing: `ala` --> requests is send, finds 999999 features and takes 5 seconds
Typing: `alab` --> request is send, finds 200 features and takes 1 second

The more precise requests is returned before the unprecise one.... and so the unprecise results will be shown after 5 seconds even if the precise result was already shown.

marcjansen:
no `fetch` then, right?

hwbllmnn:
IMO, two things would alleviate the issue:

* a delay of a few hundred ms before actually sending the search request, cancelling any pending requests that have not fired yet (not sure if this exists yet)
* augmenting the mechanism with an ordering count, discarding the responses of obsolete requests

I'll probably tackle this tomorrow, if nothing else comes up.

KaiVolland:
Fixed with #622

PR #622, opened 2018-05-04
make sure only the most recent request data is used

hwbllmnn:
### Description:

This fixes issue #621 , making sure only the most recent search request data is used. Thanks @KaiVolland !

@terrestris/devs Please review.

### DIFF
--- a/src/Field/WfsSearch/WfsSearch.jsx
+++ b/src/Field/WfsSearch/WfsSearch.jsx
@@ -144,7 +144,12 @@ export class WfsSearch extends React.Component {
      * Which prop value of option will render as content of select.
      * @type {string}
      */
-    optionLabelProp: PropTypes.string
+    optionLabelProp: PropTypes.string,
+    /**
+     * Delay in ms before actually sending requests.
+     * @type {number}
+     */
+    delay: PropTypes.number
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -291,8 +292,12 @@ export class WfsSearch extends React.Component {
     } = this.props;
 
     const request = this.getCombinedRequests();
+    const fetchTime = new Date();
 
-    this.setState({fetching: true});
+    this.setState({
+      fetching: true,
+      latestRequestTime: fetchTime.getTime()
+    });
```
dnlkoch: Not sure, if it may be relevant here, but updating the state *might* be asynchronous so the check in `onFetchSuccess()` *might* fail in succession. It might be worthwhile to check to call the fetch in the callback from `setState()`.

hwbllmnn: You're right, of course. Fixed this and the failing test in 0ff62fc.

marcjansen:
Very nice. I think it would additionally make sense to debounce the sending of the request while typing with the suggested smallish timeout. 100 ms probably? Only do thois if it can be done without hassle.

hwbllmnn:
I added a configurable timeout in 29d6b60e, with a default of 300ms (that felt right to me, might be different for others).

PR #624, opened 2018-05-04
Fix searching number fields

hwbllmnn:
### Description:

This PR fixes searching number fields via WFS. This is still far from perfect. You can configure the WFS search now with attribute details, which allows you to control how filters are created. `matchCase` can be used to control the corresponding filter attribute, `exactSearch` will make the code check for attributes of type `number` and `int` and skip the filter completely if the search term contains non numeric characters.

We should really think about doing a solr or elastic search backend in order to get better results.

@terrestris/devs Please review.

### DIFF
--- a/src/Field/WfsSearch/WfsSearch.jsx
+++ b/src/Field/WfsSearch/WfsSearch.jsx
@@ -54,7 +54,13 @@ export class WfsSearch extends React.Component {
      * An object mapping feature types to an array of attribute details.
      * @type {Object}
      */
-    attributeDetails: PropTypes.object,
+    attributeDetails: PropTypes.objectOf(PropTypes.arrayOf(
+      PropTypes.objectOf(PropTypes.shape({
+        matchCase: PropTypes.bool,
+        type: PropTypes.string,
+        exactSearch: PropTypes.bool
+      }))
+    )),
     /**
      * The namespace URI used for features.
      * @type {String}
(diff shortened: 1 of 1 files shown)


src/Field/WfsSearch/WfsSearch.jsx
```diff
@@ -50,6 +50,11 @@ export class WfsSearch extends React.Component {
      * @type {Object}
      */
     searchAttributes: PropTypes.object.isRequired,
+    /**
+     * An object mapping feature types to an array of attribute details.
+     * @type {Object}
+     */
+    attributeDetails: PropTypes.object,
```
KaiVolland: Please make use of `PropTypes.shape`.
This way you can define the existing detail keys and their types.
e.g.:
```javascript
// An object taking on a particular shape
  attributeDetails: PropTypes.shape({
    exactSearch: PropTypes.bool,
    type: PropTypes.oneOf(['int', 'string', 'number', …]),
    …
  })
```

hwbllmnn: Alright, added in 3dec246.

KaiVolland:
Nice. Didn't expect teh attribtueDetails as this complex but it looks quite strict now. :+1: 

So i guess the structure is something like this?:
```javascript
{
  layer1: [{
    attribute1: {
      matchCase: true,
      type: 'number',
      exactSearch: false
    },
    attribute2: {
      matchCase: false,
      type: 'string',
      exactSearch: true
    },
    …
  }],
  layer2: […],
  …
}
```

hwbllmnn:
Yes, exactly. It looks quite complex, but at least it matches the structure of the `searchAttributes` prop and makes accessing it easy. In an ideal world we should merge those two into a new structure, I just wanted to keep backwards compatibility.

PR #632, opened 2018-05-08
Collect coverage from src directory only

marcjansen:
### Description:

Previously we would collect coverage information from the `assets/`- and `jest/`- directories, which should not matter to us. I belive this will bring down coverage a bit, but the then reported value is a bit closer to the reality.

Update: as expected coverage went down quite a bit (`-6.8 %`).


marcjansen:
Thanks for the review.

PR #638, opened 2018-05-09
Renames FileUtil

### LINKED ISSUE #637: Rename/Fix FileUtil endings
We should rename `FileUtil.jsx` to `FileUtil.js` as `*.jsx` is for React-Classes and our util classes should not use react specific stuff.

KaiVolland:
### Description:

This changes the fileending of the `FileUtil` from `*.jsx` to `.*js`.

Fixes #637


annarieger:
Nice catch, feel free to merge :pineapple:

PR #649, opened 2018-05-14
Resolutions filter for scale combo

ahennr:
### Description:

This PR adds a prop `resolutionsFilter` to `ScaleCombo` in order to filter resolutions for which no `Select.Option` items will be created in the combo.

### DIFF
--- a/src/Field/ScaleCombo/ScaleCombo.jsx
+++ b/src/Field/ScaleCombo/ScaleCombo.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Select } from 'antd';
 const Option = Select.Option;
+import OlMap from 'ol/map';
 import {
   isInteger,
   isEmpty,
@@ -79,7 +80,7 @@ class ScaleCombo extends React.Component {
      * The map
      * @type {Ol.Map}
      */
-    map: PropTypes.object.isRequired,
+    map: PropTypes.instanceOf(OlMap).isRequired,
 
     /**
      * Set to false to not listen to the map moveend event.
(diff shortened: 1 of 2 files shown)


src/Field/ScaleCombo/ScaleCombo.jsx
```diff
@@ -193,29 +200,41 @@ class ScaleCombo extends React.Component {
...
 
-    let map = this.props.map;
     let mv = map.getView();
     // use existing resolutions array if exists
     let resolutions = mv.getResolutions();
     if (isEmpty(resolutions)) {
       for (let currentZoomLevel = mv.getMaxZoom(); currentZoomLevel >= mv.getMinZoom(); currentZoomLevel--) {
         let resolution = mv.getResolutionForZoom(currentZoomLevel);
-        this.pushScale(resolution, mv);
+        if (isFunction(resolutionsFilter)) {
```
dnlkoch: You may want to define a default resolutionsFilter like `() => true` to omit this check here (and all over).

src/Field/ScaleCombo/ScaleCombo.jsx
```diff
@@ -193,29 +200,41 @@ class ScaleCombo extends React.Component {
...
+          this.pushScale(resolution, mv);
+        }
       }
     } else {
       let reversedResolutions = reverse(clone(resolutions));
-      reversedResolutions.forEach((resolution) => {
-        this.pushScale(resolution, mv);
-      });
+      reversedResolutions
+        .filter(isFunction(resolutionsFilter) ? resolutionsFilter : f => f)
```
dnlkoch: See comment regarding default prop above.

Issue #652, opened 2018-05-16
Handle inefficiency of Webpack's Treeshaking 


amir-ba:
Since the `dist` version of `react-geo` is in `CommonJS`, `Webpack`  can't be smart enough to only load the imported parts of the code.  I would suggest that we compile the library into two folders. Where one folder uses CJS  and the other folder uses ES5 modules and in the `Package.json` we will have the `main` field point to the CJS version and the `module`  plus the `jsnext:main` point to the ES5 folder.  
This is at least what others like `Redux`  and `antd` are doing.

marcjansen:
If this is easily doable, I am plus one, but would favor feedback of more developers here.

Can we alternatively just publish Ecmascript modules and tell people how to achieve backwards compatibility in their apps? We only ever use the es modules syntax, and we have babel in in our projects, which handles the transpiling for us in our projects. @weskamm may want to weigh in here.

weskamm:
Right, the problem should also be solved by removing babel from react-geo, as its a library and not an application.
The application that uses react-geo should make use of babel, if needed.
Thoughts @dnlkoch @KaiVolland ?

amir-ba:
@marcjansen if we have both Ecmascript and CommonJS code there we also have to set the `sideEffect` tag in the `package.json` file to false. This will tell `webpack` (V4 and above) to turn:
import { a, b } from "big-module-with-flag" to import { a } from "big-module-with-flag/a"; import { b } from "big-module-with-flag/b".

dnlkoch:
Thanks @amir-ba for your well reported analysis of the issue and as far as I can see, the export of `CJS`   code and `ES6` modules looks very promising.

:+1:

marcjansen:
Hrm, is this still relevant? I don't think so.

dnlkoch:
I will close this one, but please have in mind, that we aren't publishing the ES6 modules at the moment.

PR #660, opened 2018-05-22
Bundle size reduction fixes

amir-ba:
### Description:

concerning [652](https://github.com/terrestris/react-geo/issues/652):
This PR uses the `babel-plugin-import` to reduce the size of the imported modules such as `antd`
It also changes the import patterns for turf to reduce the bundle size. 
Some libraries like `lodash` and `moment`  are structured in a way that doesn't enable them to benefit from the individual module import or the `babel-plugin-import` size reduction. 
Moreover, `ag-grid-react` just imports the complete `Ag-grid` everywhere in its code which makes the import method useless. 

@terrestris/devs  please review.


webpack.common.config.js
```diff
@@ -2,39 +2,66 @@ const path = require('path');
...
           options: {
-            javascriptEnabled: true
+            plugins:
+              [
+                [
+                  'import', [
+                    {
+                      'libraryName': 'antd',
+                      'libraryDirectory': 'es',
+                      'style': true
```
annarieger: Please check [`.babelrc`](https://github.com/terrestris/react-geo/blob/master/.babelrc) file which already contains the similar import plugin config for `antd`. One of duplicates should be removed afterwards, IMO.

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -9,15 +9,18 @@ import {
...
-  multiLineString,
   polygonize,
   polygonToLine,
-  segmentEach,
   union
 } from '@turf/turf';
-
+import {
+  getCoords,
+  multiLineString,
```
dnlkoch: Remove dangling comma.

webpack.common.config.js
```diff
@@ -2,39 +2,66 @@ const path = require('path');
...
-        'css-loader'
-      ]
-    }, {
-      test: /\.less$/,
-      use: [
-        'style-loader',
-        'css-loader',
-        {
-          loader: 'less-loader',
+    rules: [
```
dnlkoch: Please adjust indentation of the entire `rules` array.

webpack.common.config.js
dnlkoch: Good catch! 

Actually I'm pro having the config placed inside the `.babelrc` otherwise the configuration would affect `webpack` [only](https://stackoverflow.com/a/43208353).

src/Util/GeometryUtil/GeometryUtil.js
```diff
@@ -9,15 +9,18 @@ import {
...
   centroid,
   difference,
-  getCoords,
   intersect,
-  multiLineString,
   polygonize,
   polygonToLine,
-  segmentEach,
   union
 } from '@turf/turf';
```
dnlkoch: Why don't we have separate imports for the other turf modules like `@turf/buffer` here (and maybe in other modules)?

weskamm: i think this will be fixed in upcoming commit to this PR

amir-ba: yeah I am working on getting @weskamm's 
closed  <a href="https://github.com/terrestris/react-geo/pull/661">pull request</a> integrated in this PR.

webpack.common.config.js
amir-ba: I agree it makes more sense to have it in the babel setting. I will fix it.

src/Util/GeometryUtil/GeometryUtil.js
dnlkoch: :+1:

src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Modal, Input } from 'antd';
+import Modal from 'antd/lib/modal';
+import Input from 'antd/lib/input';
```
annarieger: Are you sure, that we should adjust `antd` imports by hand?

I thought, that's exactly the task of `babel-plugin-import`..

=> https://ant.design/docs/react/getting-started#Import-on-Demand

dnlkoch: And why do we import from `lib` here and from `es` in the babel config?

.babelrc
```diff
@@ -18,10 +18,15 @@
...
-        "libraryName": "antd",
-        "style": true
-      }
+      "import", [{
+          'libraryName': 'antd',
+          'libraryDirectory': 'es',
+          'style': true
+        }, {
+          'libraryName': '@turf/turf',
+          'libraryDirectory': '../',
```
dnlkoch: Please unify usage of quotes.

webpack.common.config.js
```diff
@@ -1,40 +1,44 @@
...
-      test: /\.less$/,
-      use: [
-        'style-loader',
-        'css-loader',
-        {
-          loader: 'less-loader',
-          options: {
-            javascriptEnabled: true
+        test: /\.jsx?$/,
+         exclude: /node_modules/,
```
dnlkoch: Indentation is off.

src/Button/DigitizeButton/DigitizeButton.jsx
amir-ba: we are trying to stick to an unified way of importing modules so we though why not do it for all libraries. But for some reason when importing from `antd` the plugin reduces the size much more than just the individual module import.

amir-ba: @dnlkoch because the `es` folder made some tests fail and we couldn't find the exact cause so we opted for the easiest solution. However, I would like to have the `antd` imported with the plugin since it is more effective but it is also not good to have two different patterns for different libraries.

amir-ba: yes it fixes the tests as well.

dnlkoch: No, no need for that. Just ignore it!

weskamm: To clear this up, it was my idea to have all imports harmonized and in `the right way`™

The main problems we currently face in this PR have to do with the wrong way we and other libraries use the import statements.

I know that the plugin can repair our somewhat broken kind of imports, but i thought it would be best practice to import correctly everywhere and avoid mixed imports.
Going that way would have reduced the risk of doing imports wrong in the future.

If there was enough time, it would be best to examine what exactly the difference is when using the plugin vs. explicit imports, but i guess this has to be ok for the moment ... :-)

src/HigherOrderComponent/LoadifiedComponent/LoadifiedComponent.jsx
```diff
@@ -121,7 +121,7 @@ export function loadify(WrappedComponent, {
         ...passThroughProps
       } = this.props;
 
-      // Porpoerties passed to the Spin component. For the List of all the Spin component 
+      // Porpoerties passed to the Spin component. For the List of all the Spin component
```
annarieger: Typo `Porpoerties`

src/Util/CapabilitiesUtil/CapabilitiesUtil.js
```diff
@@ -2,7 +2,7 @@ import OlWMSCapabilities from 'ol/format/wmscapabilities';
 import OlSourceImageWMS from 'ol/source/imagewms';
 import OlLayerImage from 'ol/layer/image';
 
-import { get } from 'lodash';
+import get  from 'lodash/get.js';
```
annarieger: extra whitespace after `get`

amir-ba:
@annarieger Tnx for the review.

PR #679, opened 2018-05-30
Togglebutton refactor

amir-ba:
### Description:

This PR attempts to refactor the `ToggleButton` and adds some test for it thanks to @dnlkoch. The `ToggleGroup` button is also fixed for now. But we must decide if we want to keep it at all or not.

Plz review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terrestris/react-geo",
-  "version": "8.4.0",
+  "version": "8.5.0",
   "description": "A set of geo related modules. To use in combination with react, antd and ol.",
   "license": "BSD-2-Clause",
   "author": "terrestris GmbH & Co. KG <info@terrestris.de>",
@@ -79,7 +79,7 @@
     "proj4": "2.4.4",
     "prop-types": "15.6.1",
     "query-string": "5.1.1",
-    "react-dom": "16.3.2",
+    "react-dom": "16.4.0",
     "react-dom-factories": "1.0.2",
     "react-fa": "5.0.0",
     "shpjs": "3.4.2",
--- a/src/Button/ToggleButton/ToggleButton.jsx
+++ b/src/Button/ToggleButton/ToggleButton.jsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Button, Tooltip } from 'antd';
-import Icon from 'react-fa/lib/Icon';
+import {
+  Button,
+  Tooltip
+} from 'antd';
+import { Icon } from 'react-fa';
 import isFunction from 'lodash/isFunction.js';
 
 import './ToggleButton.less';
@@ -15,7 +18,6 @@ import { CSS_PREFIX } from '../../constants';
  * @extends React.Component
  */
 class ToggleButton extends React.Component {
-
 
   /**
    * The className added to this component.
@@ -40,7 +42,7 @@ class ToggleButton extends React.Component {
     pressedIcon: PropTypes.string,
     fontIcon: PropTypes.string,
     pressed: PropTypes.bool,
-    onToggle: PropTypes.func.isRequired,
+    onToggle: PropTypes.func,
     tooltip: PropTypes.string,
     tooltipPlacement: PropTypes.string,
     className: PropTypes.string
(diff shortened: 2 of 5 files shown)


src/Button/ToggleButton/ToggleButton.jsx
```diff
@@ -94,31 +86,46 @@ class ToggleButton extends React.Component {
...
 
-    this.onClick = this.onClick.bind(this);
+    const {
+      pressed,
+      lastClickEvt
+    } = this.state;
+   
+    //   // Note: the lastClickEvt is only available if the button
+    //   // has been clicked, if the prop is changed, no click evt will
+    //   // be available.
```
KaiVolland: Double comment slashes

KaiVolland:
Could you explain the `pressed` and `propPressed`.

amir-ba:
@KaiVolland  : with the new `React` Lifecycle Methods mapping the `pressed`  `prop` to a separate internal state seems to be the only method to avoid the buggy behavior of the component. Since the new Lifecycle Methods are triggered to late for us to check if the `pressed` property Coming from outside has really changed value when compared to our internal `pressed` property.
For example in a scenario that a click has changed the internal `pressed` state if we decide to change the property of the component from outside we can't just check it against the internal `pressed` state. Therefore the variable `propPressed` only keeps track of the property changes and will trigger the button only when the property has changed form outside. The `pressed` however keeps track of the component's internal change of pressed state.

amir-ba:
@KaiVolland and @dnlkoch it seems to work now with some hacks. Can you please review. So we can make a new release where the buttons work.

src/Button/ToggleButton/ToggleButton.jsx
```diff
@@ -63,62 +65,122 @@ class ToggleButton extends React.Component {
...
+     * the following is performed here as a hack to keep track of the pressed changes.
+     * 
+     * check if the button has been clicked 
+     * |__ YES: ==> toggle the button
+     * |
+     * |__ NO: check if the prop has changed
+     *        |__ YES: ==> Toggle the button
+     *        |__ NO: check if previous update action was a click
+     *                |__ YES: ==> run the Toggle function fo the prop value
+     */
```
dnlkoch: Well documented!

src/Button/ToggleGroup/ToggleGroup.jsx
```diff
@@ -169,12 +156,8 @@ class ToggleGroup extends React.Component {
     const orientationClass = (orientation === 'vertical')
       ? 'vertical-toggle-group'
       : 'horizontal-toggle-group';
-
     const childrenWithProps = React.Children.map(children, (child) => {
-      // Only set pressed state for ToggleButtons
-      if (!(child.type === ToggleButton)) {
-        return child;
-      }
```
dnlkoch: Why did you remove this check?

src/Button/ToggleGroup/ToggleGroup.spec.jsx
```diff
@@ -103,9 +103,6 @@ describe('<ToggleGroup />', () => {
       allowDeselect: true
     });
 
-    wrapper.find(ToggleButton).first().simulate('click');
-    expect(wrapper.state().selectedName).toBe('Shinji');
```
dnlkoch: Is this not valid anymore?

package.json
```diff
@@ -79,7 +79,7 @@
     "proj4": "2.4.4",
     "prop-types": "15.6.1",
     "query-string": "5.1.1",
-    "react-dom": "16.3.2",
+    "react-dom": "16.4.0",
```
ahennr: This would be done by greenkeeper, wouldn't it?

src/Button/ToggleGroup/ToggleGroup.jsx
amir-ba: This wasn't passing the pressed property to other components that were not `Togglebutton`s like the `Digitizebutton` which is a component made of `togglebuttons` and the `pressed` property needs to gets pass through to it.

src/Button/ToggleGroup/ToggleGroup.spec.jsx
amir-ba: After changing the `allowDeselect` property it should return `null` since the button is already clicked

src/Button/ToggleButton/ToggleButton.jsx
```diff
@@ -63,62 +65,122 @@ class ToggleButton extends React.Component {
...
    *
    * @constructs ToggleButton
    */
   constructor(props) {
     super(props);
 
     // Instantiate the state.
+    // pressed: represents the pressed property of the component
+    // overallPressed: represents the pressed state that is also influenced by click events.  
+    // isClicked: shows whether the component is clicked or not
```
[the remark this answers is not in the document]
amir-ba: Now the component updates when a new prop is passed to it or when it is clicked. but it also considers the circumstances where the new property is the same as the previous but the previous click event has changed the pressed state value. This was the part where we previously had problem with.

package.json
dnlkoch: The need of this PR was (amongst others) introduced by the `react-dom` update. Since the changes made here are fixing the failing build of the [original](https://github.com/terrestris/react-geo/pull/665) greenkeeper PR, it should be fine to have it updated herein.

marcjansen:
Please clean the commit history as suggested. Otherwisw I like it and would love to see this merged.

amir-ba:
@dnlkoch , @marcjansen I got rid of the WIP commits and made it a single commit. tnx for the review.

marcjansen:
Looks good to me. Please merge.

PR #692, opened 2018-06-05
Changed find to filter to support IE

michaelsund:
### Description:
Internet explorer does not support .find without polyfills, switching to .filter()[0] produces the same result for the function getLayerByName in MapUtil.js

### DIFF
--- a/src/Util/MapUtil/MapUtil.js
+++ b/src/Util/MapUtil/MapUtil.js
@@ -168,9 +168,9 @@ export class MapUtil {
    */
   static getLayerByName(map, name) {
     const layers = MapUtil.getAllLayers(map);
-    return layers.find((layer) => {
+    return layers.filter((layer) => {
       return layer.get('name') === name;
-    });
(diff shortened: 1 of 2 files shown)


KaiVolland:
Thx @michaelsund!

You're right. I thought that babel would change this anyway but it doesn't.

I discovered another appearance of `find` in the `getQueryParam` method of the `UrlUtil`. Would be nice if you could change this too. So this PR would solve all possible problems for this topic.

michaelsund:
Ok @KaiVolland 
Do you want me to create a new PR?

KaiVolland:
I don't think this is necessary. Just commit the change into your master, as it is your source branch for this PR. The PR should update automatically.

BTW: I would recommend the usage of [feature-branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow) for the next time.

michaelsund:
Ah! will do so in the future.

PR #693, opened 2018-06-05
Adds missing bind to timeSlider onChange handler

KaiVolland:
### Description:

This adds a missing `bind` to timeSlider onChange handler.


weskamm:
nice fix, please merge

marcjansen:
A regression test would have been nice. Thanks anyhow.

PR #702, opened 2018-06-08
Several updates for the WfsSearch

dnlkoch:
### Description:

This applies some updates to the `WfsSearch` component:
  * Fix prop validation for `attributeDetails`.
  * Don't override component's `onChange` prop.
  * Don't use `optionLabelProp`, introduce `displayValue` instead.
  * Update the default `renderOption`function to include the newly introduced `displayValue` prop.
  * Fix async test to return a `Promise`.


weskamm:
nice fix for the tests, looks good to me!

dnlkoch:
Thanks for the reviews @KaiVolland and @weskamm!

PR #704, opened 2018-06-11
Fixup measure button, better mobile support

weskamm:
### Description:
This PR handles several issues with the measure buttons.
Measurements of polygons were not displayed at all on mobile devices, as the tooltips have only been set in the `pointerMoveHandler` function. This has been changed and renamed to make more sense and to get it work on mobiles.
Additionally, a `showHelpTooltip` flag has been introduced to switch the tooltips of, which is especially useful for mobile devices.


weskamm:
thx, merging

PR #705, opened 2018-06-12
Add getPathByKeyValue to ObjectUtil

dnlkoch:
### Description:

This adds the `getPathByKeyValue` method to the `ObjectUtil`, that may be used to returns the dot delimited path of a given object by the given key-value pair.

### DIFF
--- a/src/Util/ObjectUtil/ObjectUtil.js
+++ b/src/Util/ObjectUtil/ObjectUtil.js
@@ -12,6 +12,43 @@ import Logger from '../Logger';
 class ObjectUtil {
 
   /**
+   * Returns the dot delimited path of a given object by the given
+   * key-value pair. Example:
+   *
+   * ```
+   * const obj = {
+   *   level: 'first',
+   *   nested: {
+   *     level: 'second'
+   *   }
+   * };
+   * const key = 'level';
+   * const value = 'second';
+   *
+   * ObjectUtil.getPathByKeyValue(obj, key, value); // 'nested.level'
+   * ```
+   *
+   * Note: It will return the first available match!
+   *
+   * @param {Object} obj The object to obtain the path from.
+   * @param {String} key The key to look for.
+   * @param {String|Number|Boolean} value The value to look for.
+   * @param {String} [currentPath=''] The currentPath (if called in a recursion)
+   *                                  or the custom root path (default is to '').
+   */
+  static getPathByKeyValue(obj, key, value, currentPath = '') {
+    currentPath = currentPath ? `${currentPath}.` : currentPath;
+
+    for (let k in obj) {
+      if (k === key && obj[k] === value) {
+        return `${currentPath}${k}`;
+      } else if (typeof obj[k] === 'object') {
+        return ObjectUtil.getPathByKeyValue(obj[k], key, value, `${currentPath}${k}`);
+      }
+    }
+  }
+
+  /**
    * Method may be used to return a value of a given input object by a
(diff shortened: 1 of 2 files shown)


dnlkoch:
Thanks for the review @annarieger!

src/Util/ObjectUtil/ObjectUtil.js
```diff
@@ -11,6 +11,43 @@ import Logger from '../Logger';
...
+   * @param {String} [currentPath=''] The currentPath (if called in a recursion)
+   *                                  or the custom root path (default is to '').
+   */
+  static getPathByKeyValue(obj, key, value, currentPath = '') {
+    currentPath = currentPath ? `${currentPath}.` : currentPath;
+
+    for (let k in obj) {
+      if (k === key && obj[k] === value) {
+        return `${currentPath}${k}`;
+      } else if (typeof obj[k] === 'object') {
```
marcjansen: Isn't `typeof null` also `object`? Will this break the loop above?

src/Util/ObjectUtil/ObjectUtil.js
```diff
@@ -11,6 +11,43 @@ import Logger from '../Logger';
...
+   * @param {Object} obj The object to obtain the path from.
+   * @param {String} key The key to look for.
+   * @param {String|Number|Boolean} value The value to look for.
+   * @param {String} [currentPath=''] The currentPath (if called in a recursion)
+   *                                  or the custom root path (default is to '').
+   */
+  static getPathByKeyValue(obj, key, value, currentPath = '') {
+    currentPath = currentPath ? `${currentPath}.` : currentPath;
+
+    for (let k in obj) {
```
marcjansen: Will this only iterate over enumerable props? IDK, but would prefer this function to only use these keys. Maybe Object keys is an alternative. But as I said, I don't know for sure.

src/Util/ObjectUtil/ObjectUtil.js
```diff
@@ -11,6 +11,43 @@ import Logger from '../Logger';
...
+      if (k === key && obj[k] === value) {
+        return `${currentPath}${k}`;
+      } else if (typeof obj[k] === 'object') {
+        return ObjectUtil.getPathByKeyValue(obj[k], key, value, `${currentPath}${k}`);
```
marcjansen: How do we ensure this function does not run in an infinite loop, for objects that aren't serializable, because of cyclic references?

src/Util/ObjectUtil/ObjectUtil.js
dnlkoch: Yes, it will only iterater over the enumerbale props. According to [MDN Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) would result in the same behaviour since
>Object.keys() returns an array whose elements are strings corresponding to the enumerable properties found directly upon object.

src/Util/ObjectUtil/ObjectUtil.js
dnlkoch: See #710.

src/Util/ObjectUtil/ObjectUtil.js
dnlkoch: Not sure about this. At the moment I would say, that we support plain objects only…

marcjansen: Maybe we just ad a line to the docs for now

PR #710, opened 2018-06-13
Fix getPathByKeyValue to check for plain objects in recursion

dnlkoch:
### Description:
This fixes the newly introduced `getPathByKeyValue()` to only start the recursion if the given key is a plain object. It also updates the tests.


dnlkoch:
Thanks for the review (and the very useful hint) @marcjansen!

PR #724, opened 2018-06-22
Fixes the LayerTree rerendering on layerGroup update

amir-ba:
### Description:

This PR makes sure that the `LayerTree` component rerenderes when the layers in the the `layerGroup` property have changed. 

@terrestris/devs  please review.


KaiVolland:
Shouldn't the `registerAddRemoveListeners` method take care of this?

amir-ba:
@KaiVolland you are right but i don't know why it is not triggered

weskamm:
Should be fixed by #735
Closing...

PR #734, opened 2018-06-27
Fixup ToggleGroup and Buttons

weskamm:
### Description:
This fixes 2 things:
1.) The initial behaviour of a ToggleButton, when it was given a prop `pressed=true` was not handled correctly. We now call the belonging `onToggle` method to correctly initialize the button

2.) The `pressed` property of a ToggleButton now gets passed through, if it exists


KaiVolland:
Please never use the word 'fix' again with the togglebutton or -group. :rofl:

PR #772, opened 2018-07-17
[ToggleGroup] Revert pressed prop priority handling

KaiVolland:
### Description:

This fixes the `ToggleGroup` (what a classy).

For clarification:
The pressed-state of a `ToggleButton` is determined by its `pressed` prop. Which is `false` per default.
The current concept of the `ToggleGroup` is to set this via the `selectedName` which matches the `name` property of a `ToggleButton`.
  ->> As this concept doesn't seem to be very intuitive we should change it.

So the old line 165 was always true and the `ToggleButtons` never changed their state… Compare broken example.


weskamm:
Thanks for touching this, it looks ok to me, this was some oversight of mine.
ToggleGroup may need some refactoring in the future :monkey_face:

PR #793, opened 2018-08-06
Set style function with current draw style if pressed

ahennr:
### Description:

By this PR, the `getDigitizeStyleFunction` function will be applied to the `digitizeLayer` of the `DigitizeButton` is toggled. Herewith, the current set `drawStyle` is always used to draw vector features.

Plz review


ahennr:
Thx for review

PR #804, opened 2018-08-13
Fix panel pass through props

dnlkoch:
### Description:
This filters the props to be passed to the `Rnd` component correctly.

Credits to @annarieger!


annarieger:
Thanks a lot for this important fix :pineapple:

PR #834, opened 2018-08-27
make LayerTree update when prop nodeTitleRenderer has changed

jgrieb:
### Description:
The LayerTree does now update when the nodeTitleRenderer prop has changed. This is needed for example when you want to dynamically show/ hide the node title (or with image/ without image).

Please review also if the test makes sense like this.

### DIFF
--- a/src/LayerTree/LayerTree.jsx
+++ b/src/LayerTree/LayerTree.jsx
@@ -168,7 +168,7 @@ class LayerTree extends React.Component {
       }
     }
 
-    if (nodeTitleRenderer != prevProps.nodeTitleRenderer){
+    if (nodeTitleRenderer !== prevProps.nodeTitleRenderer){
       this.rebuildTreeNodes();
     }
   }
--- a/src/LayerTree/LayerTree.jsx
+++ b/src/LayerTree/LayerTree.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import isBoolean from 'lodash/isBoolean.js';
 import isFunction from 'lodash/isFunction.js';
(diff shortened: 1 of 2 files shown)


src/LayerTree/LayerTree.jsx
```diff
@@ -166,6 +167,10 @@ class LayerTree extends React.Component {
         this.rebuildTreeNodes();
       }
     }
+
+    if (nodeTitleRenderer != prevProps.nodeTitleRenderer){
```
dnlkoch: Equality check should be handled via [strict](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators) `!==` operator or lodash's `isEqual()` method.

src/LayerTree/LayerTree.spec.jsx
```diff
@@ -312,6 +312,32 @@ describe('<LayerTree />', () => {
...
+      };
+
+      const props = {
+        layerGroup,
+        map,
+        nodeTitleRenderer
+      };
+      const wrapper = TestUtil.mountComponent(LayerTree, props);
+
+      const rebuildSpy = jest.spyOn(wrapper.instance(), 'rebuildTreeNodes');
```
dnlkoch: I would suggest to create the spy before adding the prop `nodeTitleRenderer` to the component. See [here](https://github.com/terrestris/react-geo/blob/master/src/Button/ToggleButton/ToggleButton.spec.jsx#L69) for an example.

jgrieb:
Ok, I updated the code. @dnlkoch Please review if I understood your comment on changing the spyOn correctly.

Issue #846, opened 2018-08-31
AgFeatureGrid onRowClick missing original event propagation


jansule:
When using AgFeatureGrid's `onRowClick` property, two parameters will be added to the callback function: `row` and `feature`, where `row` equals the original event's `event.data` property (see code below). It would be nice if `onRowClick` would also return the original event as a whole as third property, so we don't have a loss of information here. This would be useful for instance to allow editing of specific cells but still keep the onRowClick event active.

https://github.com/terrestris/react-geo/blob/master/src/Grid/AgFeatureGrid/AgFeatureGrid.jsx#L773-L786

```javascript
// current code
onRowClick = evt => {
    const {
      onRowClick
    } = this.props;

    const row = evt.data;
    const feature = this.getFeatureFromRowKey(row.key);

    if (isFunction(onRowClick)) {
      onRowClick(row, feature);
    } else {
      this.zoomToFeatures([feature]);
    }
  }
```

```javascript
// proposed change
onRowClick = evt => {
    const {
      onRowClick
    } = this.props;

    const row = evt.data;
    const feature = this.getFeatureFromRowKey(row.key);

    if (isFunction(onRowClick)) {
      onRowClick(row, feature, evt);
    } else {
      this.zoomToFeatures([feature]);
    }
  }
```

dnlkoch:
…And the same for `onRowMouseOver`, `onRowMouseOut` and `onRowSelectionChange`. Could you also provide a fix for these methods?

PR #847, opened 2018-08-31
AgFeatureGrid add original event to onRowClick callback

jansule:
### Description:

When using AgFeatureGrid's `onRowClick` property, two parameters will be added to the callback function: `row` and `feature`, where `row` equals the original event's `event.data` property (see code below). This PR adds the original event as a whole as third property, so we don't have a loss of information here. This could be useful for instance to allow editing of specific cells but still keep the onRowClick event active.

resolves https://github.com/terrestris/react-geo/issues/846


jansule:
also added changes mentioned [here](https://github.com/terrestris/react-geo/issues/846#issuecomment-417650438)

PR #864, opened 2018-09-05
Fix window className prop to be dynamic

weskamm:
### Description:

This fixes the prop `className` to be dynamic on the window div.
Changing the prop now also changes the class, before it was only set once in the constructor.


hwbllmnn:
Nice, please merge!

PR #897, opened 2018-09-22
Before setting the extent, better cancel any animations

marcjansen:
### Description:

We should cancel pending/running animations prior to starting new ones. Otherwise erroneous behavior can be expected.

### DIFF
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
@@ -11,7 +11,7 @@ import {
 import { CSS_PREFIX } from '../../constants';
 
 /**
- * Class representing an zoom to extent button.
+ * Class representing a zoom to extent button.
  *
  *
  * @class The ZoomToExtentButton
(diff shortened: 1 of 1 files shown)


src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
```diff
@@ -40,16 +40,18 @@ class ZoomToExtentButton extends React.Component {
...
       PropTypes.instanceOf(OlSimpleGeometry)
     ]).isRequired,
 
     /**
-     * Options for fitting to the given extent. See https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit
+     * Options for fitting to the given extent. See
+     * https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit
      * @type {Object}
      */
     fitOptions: PropTypes.shape({
```
dnlkoch: Not related to the changes introduces in this PR, but the prop `fitOptions` doesn't reflect the current state of the supported options by ol. In order to become more flexible here, I would suggest to set the typing to `PropTypes.object`.

src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
```diff
@@ -40,16 +40,18 @@ class ZoomToExtentButton extends React.Component {
     map: PropTypes.instanceOf(OlMap).isRequired,
 
     /**
-     * The array extent[minx, miny, maxx, maxy]  (the values must be in the maps coordination system) or instance of Ol SimpleGeometry that the map should zoom to.
-     * @type {Array}
+     * The extent `[minx, miny, maxx, maxy]` in the maps coordinate system or an
+     * instance of Ol SimpleGeometry that the map should zoom to.
```
dnlkoch: Minor: `Ol SimpleGeometry` -> `ol.geom.SimpleGeometry`.

src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
marcjansen: :+1:

src/Button/ZoomToExtentButton/ZoomToExtentButton.jsx
marcjansen: :+1: Might change later

PR #905, opened 2018-09-26
Add handling of low-res screens in some components

jansule:
### Description:

Added media queries to a few components to make them appear smaller on low resolution screens. This  includes `ToggleButton`, `SimpleButton`, `Titlebar` and `Toolbar`. Also added a class to the modal that opens when adding a label in `DigitizeButton`.

### DIFF
--- a/src/Button/SimpleButton/SimpleButton.less
+++ b/src/Button/SimpleButton/SimpleButton.less
@@ -19,4 +19,36 @@ button.react-geo-simplebutton {
     background-color: @component-background;
     border-color: darken(@component-background, 10);
   }
+
+  @media screen and (max-width: 1280px) {
+    height: 21px;
+    .ant-btn {
+      span {
+        font-size: 10px;
+      }
+    }
+    .ant-btn-circle {
+      width: 24px;
+      height: 24px;
+
+      span {
+        font-size: 2.5em;
+        line-height: 22px;
+      }
+    }
+  }
+}
+
+button.react-geo-togglebutton {
+  @media screen and (max-width: 1280px) {
+    &.ant-btn-circle {
+      width: 24px;
+      height: 24px;
+
+      span {
+        font-size: 2.5em;
+        line-height: 22px;
+      }
+    }
+  }
 }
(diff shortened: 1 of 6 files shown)


src/Button/SimpleButton/SimpleButton.less
```diff
@@ -19,4 +19,36 @@ button.react-geo-simplebutton {
...
+
+      span {
+        font-size: 2.5em;
+        line-height: 22px;
+      }
+    }
+  }
+}
+
+button.react-geo-togglebutton {
```
dnlkoch: The same definition is also available in the `ToggleButton.less`. Is this intended?

src/Button/SimpleButton/SimpleButton.less
```diff
@@ -19,4 +19,36 @@ button.react-geo-simplebutton {
     background-color: @component-background;
     border-color: darken(@component-background, 10);
   }
+
+  @media screen and (max-width: 1280px) {
```
dnlkoch: The threshold value of `1280px` could possibly be moved to a less variable (see [here](https://github.com/terrestris/react-geo/blob/master/src/style/variables.less)).

src/Button/SimpleButton/SimpleButton.less
jansule: No. Thanks for the hint. Copy&Paste error. Will be resolved with next commit.

marcjansen:
Can you provide some screenshots before and after? On those screens? Thx

marcjansen:
Also, does this somehow clash with the widths of some popular smartphones? I.e. is mobile display equally good before and after?

jansule:
@ahennr and I took a look at the quality on smartphone screens. They are affected by this change, but still working fine. Buttons are still large enough.

jansule:
**Before:**
image

**After:**
image

weskamm:
Usually we want bigger buttons on mobiles, as they are harder to use when too small.
Not sure if i am mixing things up?

jansule:
Talked to @weskamm and @ahennr and we came to the conclusion that it does not make sense to make these changes directly in react-geo but rather within the respective projects. Also as @weskamm pointed out, reducing the button sizes is counterproductive for mobile development.

Closing PR.

PR #906, opened 2018-09-27
Provide optional callback prop `onFeatureSelect` to be used with `ol.interaction.Select` (ol4)

annarieger:
### Description:
Title says it all.

Possible use case: `styleFunction` of `ol.select.Interaction` doesn't take effect when selected feature already has style. The proposed callback function can be used to force setting of overriden select style.


annarieger:
Since #907 containing similar changeset for `master` branch is approved and merged merging this also now..

PR #908, opened 2018-09-27
Enhance label handling in `DigitizeButton` component (ol4)

annarieger:
### Description:
* Use `ant` `Input.TextArea` instead of `Input` to enable multiline labels
* Set currently modifying label as value in label edit modal (previously value was always empty)
* Introduced new numerical prop `maxLabelLineLength` to determine if label should be divided into multiline label after provided threshold is exceeded
* Ensure that `onModalLabelOk` / `onModalLabelCancel` callback function will be called with updated values by moving them into `setState` callback

### DIFF
--- a/src/Button/DigitizeButton/DigitizeButton.jsx
+++ b/src/Button/DigitizeButton/DigitizeButton.jsx
@@ -1004,10 +1004,7 @@ class DigitizeButton extends React.Component {
     this.setState({
       showLabelPrompt: false
     }, () => {
-      this.setTextOnFeature(this._digitizeTextFeature);
-      if (onModalLabelOk) {
-        onModalLabelOk(this._digitizeTextFeature, textLabel);
-      }
+      this.setTextOnFeature(this._digitizeTextFeature, onModalLabelOk);
     });
   }
 
@@ -1029,7 +1026,7 @@ class DigitizeButton extends React.Component {
         this._digitizeFeatures.remove(this._digitizeTextFeature);
         this._digitizeTextFeature = null;
       }
-      if (onModalLabelCancel) {
+      if (isFunction(onModalLabelCancel)) {
         onModalLabelCancel();
       }
     });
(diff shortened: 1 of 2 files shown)


src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -897,7 +907,8 @@ class DigitizeButton extends React.Component {
     if (feature.get('isLabel')) {
       this._digitizeTextFeature = feature;
       this.setState({
-        showLabelPrompt: true
+        showLabelPrompt: true,
+        textLabel: feature.getStyle().getText().getText() || ''
```
dnlkoch: Can `feature.getStyle().getText()` become undefined here?

src/Button/DigitizeButton/DigitizeButton.jsx
```diff
@@ -1018,22 +1022,37 @@ class DigitizeButton extends React.Component {
...
-    const label = StringUtil.stringDivider(this.state.textLabel, 16, '\n');
+  setTextOnFeature = (feat, onModalOkCbk) => {
+    const {
+      maxLabelLineLength
+    } = this.props;
+
+    let label = this.state.textLabel;
+    if (maxLabelLineLength) {
+      label = StringUtil.stringDivider(
+        this.state.textLabel, maxLabelLineLength, '\n'
```
dnlkoch: Haven't checked, what `stringDivider` does in detail, but I would suggest to pass a cloned or destructured `textLabel` string to the method to ensure, the state object won't be manipulated in it.

src/Button/DigitizeButton/DigitizeButton.jsx
annarieger: No, since we check above if feature has property `isLabel`, which is extra set on labeled features only

annarieger:
Thanks for review, merging :pineapple:

PR #910, opened 2018-09-27
Enhance label handling in `DigitizeButton` component (ol5)

annarieger:
### Description:
* Use `ant` `Input.TextArea` instead of `Input` to enable multiline labels
* Set currently modifying label as value in label edit modal (previously value was always empty)
* Introduced new numerical prop `maxLabelLineLength` to determine if label should be divided into multiline label after provided threshold is exceeded
* Ensure that `onModalLabelOk` / `onModalLabelCancel` callback function will be called with updated values by moving them into `setState` callback

ol4 compatible PR version can be found under #908.


annarieger:
Comments are addressed in the same way as in #908.

Thanks for review, merging :pineapple:

PR #963, opened 2018-12-18
Introduces onClear prop to NominatimSearch

KaiVolland:
### Description:

This introduces an `onClear`method to the `NominatimSearch` which can be passed as a prop.

### DIFF
--- a/src/Field/NominatimSearch/NominatimSearch.jsx
+++ b/src/Field/NominatimSearch/NominatimSearch.jsx
@@ -102,6 +102,12 @@ export class NominatimSearch extends React.Component {
      */
     onSelect: PropTypes.func,
     /**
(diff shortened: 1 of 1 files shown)


src/Field/NominatimSearch/NominatimSearch.jsx
```diff
@@ -200,6 +210,10 @@ export class NominatimSearch extends React.Component {
         this.doSearch();
       }
     });
+
+    if (!inputValue && onClear) {
```
dnlkoch: Note: This will evaluate the input value `0` as truthy. Not sure if this is intended.

KaiVolland: Shouldn't be a problem as the inputValue is of type `string` or `undefiend`.

PR #1001, opened 2019-01-28
Upgrade to babel 7

### LINKED ISSUE #989: Update to babel 7
We should update to babel version 7.

See #978 as well.

dnlkoch:
### Description:
This upgrades `Babel` (and all related packages like the presets and plugins) to version 7, `babel-loader` to version 8 and `Jest` to version 24.

Fixes #989.


dnlkoch:
Actually this shouldn't be necessary as the `base-util` has been rewritten to TypeScript already and the `ol-util` will follow sooner or later as well.

PR #1032, opened 2019-02-15
Forces rebuildTreeNodes onExpand

KaiVolland:
### Description:

This introduces a `onExpand` listener that calls `rebuildTreeNodes`, as expanding Trees with more the 1 level didn't work properly.


hwbllmnn:
Nice one, this annoyed me previously...

PR #1043, opened 2019-02-19
Fix `onEscape` function call for Window component

annarieger:
### Description:
`Window` component  won't be autofocused on render. That leads to missing behaviour of `onEscape` function which can be optionally provided as prop (e.g. closing window on Escape didn't work anymore out of the box -> to get it work, user had to click manually into the window body firstly to set focus on component).
This PR fix the described issue, `onEscape` handler will be called now even if the Window component is not currently focused.

### DIFF
--- a/src/Panel/Panel/Panel.jsx
+++ b/src/Panel/Panel/Panel.jsx
@@ -202,13 +202,22 @@ export class Panel extends React.Component {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


weskamm:
:+1:

annarieger:
Thanks! :pineapple:

PR #1046, opened 2019-02-20
fix test for onKeyDown method

annarieger:
### Description:
Follow up of #1043, fixes test which was ok for travis but fails locally.

Please kindly review @weskamm


weskamm:
nice :+1:

annarieger:
Thanks again, merging :pineapple:

PR #1053, opened 2019-02-22
Fix escape key for IE

annarieger:
### Description:
Tiny fix to ensure that `Escape` key will also be matched while using IE.

### DIFF
--- a/src/Panel/Panel/Panel.jsx
+++ b/src/Panel/Panel/Panel.jsx
@@ -34,7 +34,7 @@ export class Panel extends React.Component {
     * @type {String}
     * @private
     */
-   _escapeKeyboardEventKey = 'Escape';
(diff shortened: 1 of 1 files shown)


marcjansen:
How about using `evt.keyCode`? It feels as if this is better for cross browser compatibility, I have no proof though.

https://keycode.info/ & https://developer.mozilla.org/de/docs/Web/API/KeyboardEvent/keyCode have insides, can be tested in the various browsers.

annarieger:
`evtCode` is deprecated --> https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

marcjansen:
Fine for me, even though support for `key` isn't so great as well: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility

PR #1083, opened 2019-03-20
Fixup nominatim search by comparing strings

weskamm:
### Description:
This fixes the `onMenuItemSelected` handler of the `NominatimSearch`, which is currently broken due to a change in the response of the API. 

The property `place_id` has been changed from string to a number, which causes a comparison to fail. This has been fixed by comparing both values stringified with `toString()`

### DIFF
--- a/src/Field/NominatimSearch/NominatimSearch.jsx
+++ b/src/Field/NominatimSearch/NominatimSearch.jsx
@@ -267,9 +267,9 @@ export class NominatimSearch extends React.Component {
    * @param {value} key The key of the selected option.
    */
   onMenuItemSelected(key) {
-    const selected = this.state.dataSource.filter(
+    const selected = this.state.dataSource.find(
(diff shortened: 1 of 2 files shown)


src/Field/NominatimSearch/NominatimSearch.jsx
```diff
@@ -267,7 +267,9 @@ export class NominatimSearch extends React.Component {
    * @param {value} key The key of the selected option.
    */
   onMenuItemSelected(key) {
-    const selected = this.state.dataSource.filter(i => i.place_id === key)[0];
+    const selected = this.state.dataSource.filter(
+      i => i.place_id.toString() === key.toString()
+    )[0];
```
hwbllmnn: You can also use `.find` here which just returns the first match.

weskamm: changed it

Issue #1097, opened 2019-04-01
WfsSearch define id column


gkvoelkl:
In WfsSearch I can define the displayed value

` displayValue='LABEL_TEXT' `

I there a way to define the name of the id column ?

At the moment it is 'id'

marcjansen:
If there isn't a way, we probably should have one... @kaivolland or @dnlkoch, can you shed a light?

KaiVolland:
This is currently not possible. I'll add it.

dnlkoch:
Thanks @gkvoelkl for posting this issue, it is fixed and is now available via version [11.3.0](https://github.com/terrestris/react-geo/releases/tag/v11.3.0).

marcjansen:
Thanks everybody!

PR #1098, opened 2019-04-02
Customizable id property in WfsSearch

dnlkoch:
### Description:
This adds the prop `idProperty` to the `WfsSearch` to define a custom id property for the requested feature types.

See #1097.


dnlkoch:
Thanks for the reviews!

PR #1120, opened 2019-05-07
Update README

dnlkoch:
### Description:

Adds development requirements to README. npm 6.8.0 is needed for release via np (since > 5).


marcjansen:
Thanks for this. Can we additionally use the `engines` field from `package.json`?

PR #1191, opened 2019-06-21
Update gh-pages branch if tag has been created

dnlkoch:
### Description:
This fixes the update of the `gh-pages` branch to create a versioned snapshot of the styleguide if a tag has beed created.


dnlkoch:
Thx for the review @annarieger!

PR #1252, opened 2019-11-20
Introduces LayerSwitcher

KaiVolland:
### Description:

This introduces the `LayerSwitcher`.
A component designed to be used for switching between two (or more) backgroundlayers.

The view of the `LayerSwitcher` is synched with the maps view.

localhost_9090_portal_tools=identify-connect,organise-implement

### DIFF
--- a/src/LayerSwitcher/LayerSwitcher.less
+++ b/src/LayerSwitcher/LayerSwitcher.less
@@ -19,11 +19,17 @@
   }
 
   .layer-title {
+    font-size: 0.8em;
     position: absolute;
     width: 100%;
     bottom: 0;
-    padding: 3px;
+    left: 0;
+    padding: 3px 6px;
+    border-radius: 0 0 4px 4px;
     color: white;
-    background-image: linear-gradient(transparent,rgba(0,0,0,0.6));
+    background-image: linear-gradient(rgba(0,0,0,0.01),rgba(0,0,0,0.7));
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
 }
--- a/src/LayerSwitcher/LayerSwitcher.less
+++ b/src/LayerSwitcher/LayerSwitcher.less
@@ -1,0 +1,29 @@
+.react-geo-layer-switcher {
+  .clip {
+    width: 80px;
+    height: 80px;
+    background-color: white;
+    border: 2px solid black;
+    border-radius: 4px;
+    overflow: hidden;
+    cursor: pointer;
+  }
+
+  #layer-switcher-map {
+    width: 256px;
+    height: 256px;
+    position: relative;
+    left: -68px;
+    top: -68px;
+    pointer-events: none;
+  }
+
+  .layer-title {
+    position: absolute;
+    width: 100%;
+    bottom: 0;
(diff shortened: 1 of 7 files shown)


src/LayerSwitcher/LayerSwitcher.tsx
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /**
+   * Clones a layer
+   *
+   * @param layer The layer to clone.
+   * @returns The cloned layer.
+   */
+  cloneLayer = (layer: OlLayerTile | OlLayerGroup) => {
+    let layerClone: OlLayerTile | OlLayerGroup;
```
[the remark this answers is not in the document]
KaiVolland: As the cloneLayer method needs a real class to construct the i tried to keep it simple for the first approach. We might want to add a more complex `cloneLayer` method to the ol-util package later.

src/LayerSwitcher/LayerSwitcher.less
```diff
@@ -0,0 +1,29 @@
...
+    pointer-events: none;
+  }
+
+  .layer-title {
+    position: absolute;
+    width: 100%;
+    bottom: 0;
+    padding: 3px;
+    color: white;
+    background-image: linear-gradient(transparent,rgba(0,0,0,0.6));
```
marcjansen: How does it look like if the title is too long? Will it be clipped? Can we use e.g. text-shadow to have the title be readable regardless of the actual layer?

src/LayerSwitcher/LayerSwitcher.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const {
+      className
+    } = this.props;
+
+    const finalClassName = className
+      ? `${className} ${this.className}`
+      : this.className;
+
+    return (
+      <div className={finalClassName}>
```
marcjansen: Can we have some sort of a hint to users how to interact with the component? like a change of cursor, a tooltip like "Click to switch on the Sentinel-2 layer, 3 more layers available" or sth. like this?

src/LayerSwitcher/LayerSwitcher.tsx
```diff
@@ -0,0 +1,259 @@
...
+   *
+   */
+  onSwitcherClick = (evt: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+    evt.stopPropagation();
+    this._map.getLayers().getArray().forEach((layer, index) => {
+      if (layer.getVisible()) {
+        this._visibleLayerIndex = index;
+      }
+    });
+    this.updateLayerVisibilty();
```
marcjansen: Misspelled?

src/LayerSwitcher/LayerSwitcher.less
KaiVolland: text-overflow is not yet tested. I will add it.

The background-image gradient is there to ensure the readability of the the title.

src/LayerSwitcher/LayerSwitcher.tsx
KaiVolland: I don't think a tooltip is necessary. If you need one in your application you can still wrapp it around the component.

PR #1256, opened 2019-11-21
TypeScript fixes / changes

KaiVolland:
### Description:

This fixes several typings of the components.

:exclamation: This also includes a small refactoring of the `DigitizeButton`. Please consider when reviewing.


src/Button/DigitizeButton/DigitizeButton.tsx
```diff
@@ -429,7 +434,6 @@ class DigitizeButton extends React.Component<DigitizeButtonProps, DigitizeButton
 
     this.state = {
       digitizeLayer: null,
```
dnlkoch: The digitize layer should be ported to a private class member as well.

PR #1264, opened 2019-12-06
Introduces the TimeLayerSliderPanel for WMS-T layer

mholthausen:
# FEATURE

## Description

This merges the existing `TimeSlider` and `TimeLayerAware` components into the new panel component `TimeLayerSliderPanel`.

timelayersliderpanel

Some points are still open:

* [x] [tooltips](https://github.com/terrestris/react-geo/compare/master...mholthausen:introduce-timelayersliderpanel?expand=1#diff-0e03078ed2beb0b1d78fb55d39828718R476): object access via string literals is disallowed
* [x] [ts-ignore](https://github.com/terrestris/react-geo/compare/master...mholthausen:introduce-timelayersliderpanel?expand=1#diff-0e03078ed2beb0b1d78fb55d39828718R183): When removed, the application fails to compile `Argument of type 'typeof TimeSlider' is not assignable to parameter of type 'ComponentType<TimeSliderProps>'.`
* [x] Tests: no working tests included until yet

Maybe another rect-geo @terrestris/devs can help me out here soon.

### DIFF
--- a/src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
+++ b/src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 
 import moment from 'moment';
+import OlLayer from 'ol/layer/Layer';
+import OlMap from 'ol/Map';
+// import _get from 'lodash/get';
 
 const _isFinite = require('lodash/isFinite');
 const _isEqual = require('lodash/isEqual');
@@ -34,20 +37,19 @@ export interface Tooltips {
   dataRange: string;
 }
 
+export type PlaybackSpeedType = 'hours' | 'days' | 'weeks' | 'months' | 'years';
+
 export interface DefaultTimeLayerSliderPanelProps {
   className: string;
   onChange: (arg: moment.Moment) => void;
-  import OlLayer from 'ol/layer/OlLayer';
-  (…)
   timeAwareLayers: OlLayer[];
   value: moment.Moment;
   dateFormat: string;
   tooltips: Tooltips;
+  autoPlaySpeedOptions: number[];
 }
 
 export interface TimeLayerSliderPanelProps extends Partial<DefaultTimeLayerSliderPanelProps> {
-  import OlMap from 'ol/Map';
-  (…)
   map: OlMap;
   initStartDate: moment.Moment;
   initEndDate: moment.Moment;
@@ -55,18 +57,11 @@ export interface TimeLayerSliderPanelProps extends Partial<DefaultTimeLayerSlide
 
 export interface TimeLayerSliderPanelState {
   value: moment.Moment;
-  playbackSpeed: string;
+  playbackSpeed: number | PlaybackSpeedType;
   autoPlayActive: boolean;
   startDate: moment.Moment;
   endDate: moment.Moment;
 }
-
-/**
- * mapStateToProps - mapping state to props of Map Component.
- *
- * @param {Object} state current state
- * @return {Object} mapped props
- */
 
 /**
  * The panel combining all time slider related parts.
@@ -85,7 +80,7 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
    */
   public static defaultProps: DefaultTimeLayerSliderPanelProps = {
     className: '',
-    onChange: () => {/* empty */},
+    onChange: () => {},
     timeAwareLayers: [],
     value: moment(moment.now()),
     dateFormat: 'YYYY-MM-DD',
@@ -97,7 +92,8 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
       months: 'Months',
       years: 'Years',
       dataRange: 'Set data range'
-    }
+    },
+    autoPlaySpeedOptions: [ 0.5, 1, 2, 5, 10, 100, 300 ]
   };
 
   /**
@@ -108,7 +104,7 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
 
     this.state = {
       value: moment().milliseconds(0),
-      playbackSpeed: '1',
+      playbackSpeed: 1,
       autoPlayActive: false,
       startDate: moment().milliseconds(0),
       endDate: moment().milliseconds(0),
@@ -136,12 +132,14 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
   }
 
   componentDidUpdate(prevProps: TimeLayerSliderPanelProps) {
-    // TODO: this deep check may impact performance..
-    if (!(_isEqual(prevProps.timeAwareLayers, this.props.timeAwareLayers))) {
-      // update slider properties if layers were updated
-      this.wrapTimeSlider();
-      this.findRangeForLayers();
-    }
+    prevProps.timeAwareLayers.forEach((pl: any, i: number) => {
+      const tpl = this.props.timeAwareLayers[i];
+      if (!(_isEqual(pl.ol_uid, tpl.ol_uid))) {
+        // update slider properties if layers were updated
+        this.wrapTimeSlider();
+        this.findRangeForLayers();
+      }
+    });
   }
 
   /**
@@ -154,10 +152,11 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
       value,
       autoPlayActive,
       startDate,
-      endDate
+      endDate,
     } = this.state;
     const {
-      timeAwareLayers
+      timeAwareLayers,
+      tooltips
     } = this.props;
 
     if (nextState.value !== value) {
@@ -173,6 +172,9 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
       return true;
     }
     if (!(_isEqual(nextProps.timeAwareLayers, timeAwareLayers))) {
+      return true;
+    }
+    if (!(_isEqual(nextProps.tooltips, tooltips))) {
       return true;
     }
     return false;
@@ -242,7 +244,7 @@ export class TimeLayerSliderPanel extends React.Component<TimeLayerSliderPanelPr
   /**
    * Handler for the time slider change behaviour
    */
-  timeSliderCustomHandler = (value: any) => {
+  timeSliderCustomHandler = (value: object) => {
     const currentMoment = moment(value).milliseconds(0);
     const newValue = currentMoment.clone();
     this.setState({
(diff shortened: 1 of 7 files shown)


src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+
+export interface TimeLayerSliderPanelProps extends Partial<DefaultTimeLayerSliderPanelProps> {
+  map: any;
+  initStartDate: moment.Moment;
+  initEndDate: moment.Moment;
+}
+
+export interface TimeLayerSliderPanelState {
+  value: moment.Moment;
+  playbackSpeed: string;
```
dnlkoch: I would expect a `number` here.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+  startDate: moment.Moment;
+  endDate: moment.Moment;
+}
+
+/**
+ * mapStateToProps - mapping state to props of Map Component.
+ *
+ * @param {Object} state current state
+ * @return {Object} mapped props
+ */
```
dnlkoch: Please remove unneeded multiline comment.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+  /**
+   * The default props of LayerSetBaseMapChooser
+   *
+   * @static
+   * @type {DefaultLayerSetBaseMapChooserProps}
+   * @memberof LayerSetBaseMapChooser
+   */
+  public static defaultProps: DefaultTimeLayerSliderPanelProps = {
+    className: '',
+    onChange: () => {/* empty */},
```
dnlkoch:
```suggestion
    onChange: () => {},
```

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+
+    this.setState({
+      startDate: initStartDate,
+      endDate: initEndDate
+    });
+  }
+
+  componentDidUpdate(prevProps: TimeLayerSliderPanelProps) {
+    // TODO: this deep check may impact performance..
+    if (!(_isEqual(prevProps.timeAwareLayers, this.props.timeAwareLayers))) {
```
dnlkoch: You may want to compare the `ol_uid` of each layer instead.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+      // update slider properties if some another layer set was chosen
```
dnlkoch:
```suggestion
      // update slider properties if layers were updated
```

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+    if (nextState.startDate !== startDate) {
+      return true;
+    }
+    if (nextState.endDate !== endDate) {
+      return true;
+    }
+    if (!(_isEqual(nextProps.timeAwareLayers, timeAwareLayers))) {
+      return true;
+    }
+    return false;
```
dnlkoch: I like the approach, but please also add update support for at least the `tooltips` prop.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+    if (this.props.onChange) {
+      this.props.onChange(newValue);
+    }
+  }
+
+  /**
+   * makes sure that the appended time parameter in GetMap calls
+   * is rounded to full hours to receive a valid response
+   */
+  wmsTimeHandler = (value?: any) => {
```
dnlkoch: Why do we need this? A WMS-Time may also accept a non-full-hours time parameter, see e.g. [here](https://docs.geoserver.org/stable/en/user/services/wms/time.html).

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+        transform: 'translate(50%)'
+      }
+    };
+    marks[mid.toISOString()] = {
+      label: mid.format(dateFormat)
+    };
+
+    return (
+      <div className={`time-layer-slider ${disabledCls}`.trim()}>
+
```
dnlkoch: Please remove empty line.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+          className={extraCls + ' speed-picker'}
+          onChange={this.onPlaybackSpeedChange}
+        >
+          <Option value="0.5">0.5x</Option>
+          <Option value="1">1x</Option>
+          <Option value="2">2x</Option>
+          <Option value="5">5x</Option>
+          <Option value="10">10x</Option>
+          <Option value="100">100x</Option>
+          <Option value="300">300x</Option>
```
dnlkoch: It would be nice to have this list configurable.

src/Slider/TimeSlider/TimeSlider.tsx
```diff
@@ -52,7 +52,7 @@ export interface BaseProps {
   marks?: SliderMarks;
 }
 
-export type TimeSliderProps = BaseProps & Partial<DefaultProps> & SliderProps;
```
dnlkoch: Having the `TimeSliderProps` extend the `SliderProps` seems very legit to me, why did you remove them?

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
KaiVolland: … also remove `{Object}`. The types are determined by TypeScript now.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+
+  private _TimeLayerAwareSlider: any;
+  private _wmsTimeLayers: any[];
+  private _interval: number;
+
+  /**
+   * The default props of LayerSetBaseMapChooser
+   *
+   * @static
+   * @type {DefaultLayerSetBaseMapChooserProps}
```
KaiVolland:
```suggestion
```

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
```diff
@@ -0,0 +1,496 @@
...
+    });
+    newStartDate = startDatesFromLayers.length > 0 ? moment.min(startDatesFromLayers) : startDate;
+    newEndDate = endDatesFromLayers.length > 0 ? moment.max(endDatesFromLayers) : endDate;
+    this.updateDataRange([newStartDate, newEndDate]);
+  }
+
+  /**
+   * Handler for the time slider change behaviour
+   */
+  timeSliderCustomHandler = (value: any) => {
```
KaiVolland: The type of the SliderHandler should be known. Maybe an Event or number array.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
KaiVolland: I think it should be `onChange: (value: moment.Moment) => void,`

KaiVolland: If it is empty it shouldn't be a `defaultProp`

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: Th playback speed can also be set to `moment()` values like `'hours'`, `'days'`, `'months'` etc. Since these are then passed as `string`, the entire data type of this variable has probably been set to the type of `string`.

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: If it is set like suggested, the linting fails:` no-empty: block is empty`. The proposed solutions was to enter an comment into this block.
If these are removed completely, compilation fails, because this information is missing in `TimeLayerAware`. Thus, I would suggest to leave them as they are, or do you have another suggestion @KaiVolland ?

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: Comparing the `ol_uid` of each layer in `timeAwareLayers` of `prevProps` against the layer at the same position of `this.props`

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: I don't know, it was implemented that way in the react-geo-baseclient. It may have been needed to display images for a timespan of a complete hour, if there is only one image available for a solid hour. I have now added a query to the configuration of the layer, whether this behavior is desired or not (`roundToFullHours`).

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: `autoPlaySpeedOptions` can be passed to the panel now. If these are not specified, default values are used.

src/Slider/TimeSlider/TimeSlider.tsx
mholthausen: We have removed these, because otherwise an error occurs with `timeLayerAware(TimeSlider, this._wmsTimeLayers);`: `Argument of type 'typeof TimeSlider' is not assignable to parameter of type 'ComponentType<TimeSliderProps>'`. This could only be bypassed for compilation with the specification `\\ @ts-ignore`

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.tsx
mholthausen: Set the `tslint.json` to allow empty functions.

src/Slider/TimeSlider/TimeSlider.tsx
mholthausen: The doubled props defined by the antd slider are omitted.

PR #1509, opened 2020-06-26
Update antd to version 4

dnlkoch:
### Description:

This updates `antd` to [version 4](https://ant.design/docs/react/migration-v4). All components except the `LayerTree` have been updated to the updated API. Migrating the `LayerTree` to make use of the `treeData` prop for generating the `LayerTreeNode`s may require a complete rewrite of the component which is out of the goal here.

## Breaking changes

* peerDependency requires `antd` in version 4.
* Removed property `fontIcon` from the `SimpleButton` .
* Property `icon` in the `SimpleButton` has been changed to accept a `ReactNode` instead of a `string`.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -21,7 +21,7 @@ module.exports = {
     '<rootDir>/dist/'
   ],
   transformIgnorePatterns: [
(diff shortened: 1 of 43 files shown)


dnlkoch:
Thanks for the review!

PR #1649, opened 2020-08-21
MapContext and useMap hook

KaiVolland:
### Description:

This introduces the `MapContext` and the `useMap` hook.

It is a replacement for the `MapProvider` (which is now marked as deprecated) and the `mappify` HOC.


KaiVolland:
Hey @marcjansen thanks for the comments. Answers are inline.

> A couple of questions & comments
> 
> * Why is this change helpful? Can we link to resources elsewhere on the web?

React 16.3 introduced a new Context-API https://reactjs.org/docs/context.html. And with 16.8 hooks where introduced https://reactjs.org/docs/hooks-intro.html.
These are two new features which are quite cool. Especially when it comes to Syntax.
> * Why do we deprecate and not remove entirely?

I think deprecating is a good option to keep projects with the old MapProvider running when updating react-geo. The change wouldn't be that big, but i like the idea of logging a warning that things have changed and could be done cooler now.
> * Is it for users clear how to migrate their usage of the deprecated to the now recommended approach?

Yes, the examples and tests should be enough to understand how things have changed.
> * We should have issues in the workshop repos highlighting the needed changes. If I understand it correctly, the changes should be doable also by not-so-experienced users, sth. for Hacktoerfest even?

:+1:
> * how long do we intend to keep the deprecated functionality in the code?

I didn't thought much about this point.  
> * Can we please start creating [changelogs](https://keepachangelog.com/de/1.0.0/)?

We do have the release notes https://github.com/terrestris/react-geo/releases which is more or less what you are looking for.
If you'd prefere to have this in a single file please open an issue.

src/Map/MapComponent/MapComponent.example.md
```diff
@@ -2,6 +2,8 @@ This example shows the usage of the MapComponent in combination with the MapProv
 It makes use of the `mappify` HOC function to supply the provided map to the MapComponent
 and the NominatimSearch.
 
+**THE HOC IS DEPREACTED. PLEASE CONSIDER USING THE MAPCONTEXT WITH `useMap`**
```
dnlkoch, line 5:
```suggestion
**THE HOC IS DEPRECATED AND ONLY AVAILABLE FOR BACKWARDS COMPATIBILITY. PLEASE CONSIDER USING THE MAPCONTEXT WITH `useMap`**
```

dnlkoch (approved):

dnlkoch:
Thanks @KaiVolland for the integration this hook! Absolutely helpful!

@marcjansen some additinal thoughts about questions:

> Why is this change helpful? Can we link to resources elsewhere on the web?
> Why do we deprecate and not remove entirely?

React recommends the usage of function components for new projects/applications, but for existing apps, which make use of react components, the HOC approach should still exist for backwards compatibility. 

> Is it for users clear how to migrate their usage of the deprecated to the now recommended approach?

I think the examples make this very clear. Also the usage of hooks is very common right now.

> We should have issues in the workshop repos highlighting the needed changes. If I understand it correctly, the changes should be doable also by not-so-experienced users, sth. for Hacktoerfest even?

The workshop makes use of function components and hooks already, but it's a good point to add the proposed change to it. Could you please open an issue in the react-geo-ws?

> how long do we intend to keep the deprecated functionality in the code?

The HOC should at least stay alive until the react components are officially supported.

> Can we please start creating changelogs?

Actually I like the current way we handle the release notes. As suggested by @KaiVolland please open an issue open for discussion if you miss anything.

marcjansen:
> Hey @marcjansen thanks for the comments. Answers are inline.
> 
> > A couple of questions & comments
> > 
> > * Why is this change helpful? Can we link to resources elsewhere on the web?
> 
> React 16.3 introduced a new Context-API https://reactjs.org/docs/context.html. And with 16.8 hooks where introduced https://reactjs.org/docs/hooks-intro.html.
> These are two new features which are quite cool. Especially when it comes to Syntax.

Cool, this info should be in example, if it isn't already.

> 
> > * Why do we deprecate and not remove entirely?
> 
> I think deprecating is a good option to keep projects with the old MapProvider running when updating react-geo. The change wouldn't be that big, but i like the idea of logging a warning that things have changed and could be done cooler now.
> 

Hmm. Does deprecation also mean that we will not be fixing bugs in the deprecated functionality? Otherwise we would be supporting features we do not really want, right?

> > * Is it for users clear how to migrate their usage of the deprecated to the now recommended approach?
> 
> Yes, the examples and tests should be enough to understand how things have changed.

Nice, glad you say that.

> 
> > * We should have issues in the workshop repos highlighting the needed changes. If I understand it correctly, the changes should be doable also by not-so-experienced users, sth. for Hacktoerfest even?
> 
> 

Will you be creating some?

> 
> > * how long do we intend to keep the deprecated functionality in the code?
> 
> I didn't thought much about this point.

I think @dkoch said for as long as they are supported by React, right? That can be quite the time...

> 
> > * Can we please start creating [changelogs](https://keepachangelog.com/de/1.0.0/)?
> 
> We do have the release notes https://github.com/terrestris/react-geo/releases which is more or less what you are looking for.
> If you'd prefere to have this in a single file please open an issue.

I just did: https://github.com/terrestris/react-geo/issues/1652

Thanks for the answers.

marcjansen:
> Thanks @KaiVolland for the integration this hook! Absolutely helpful!
> 
> @marcjansen some additinal thoughts about questions:
> 
> > Why is this change helpful? Can we link to resources elsewhere on the web?
> > Why do we deprecate and not remove entirely?
> 
> React recommends the usage of function components for new projects/applications, but for existing apps, which make use of react components, the HOC approach should still exist for backwards compatibility.


Yeah, but they simply wouldn't be able to upgrade, right? Which is a sort of nice pressure to stay in sync with the latest.


> 
> > Is it for users clear how to migrate their usage of the deprecated to the now recommended approach?
> 
> I think the examples make this very clear. Also the usage of hooks is very common right now.
> 
> > We should have issues in the workshop repos highlighting the needed changes. If I understand it correctly, the changes should be doable also by not-so-experienced users, sth. for Hacktoerfest even?
> 
> The workshop makes use of function components and hooks already, but it's a good point to add the proposed change to it. Could you please open an issue in the react-geo-ws?
> 
> > how long do we intend to keep the deprecated functionality in the code?
> 
> The HOC should at least stay alive until the react components are officially supported.
> 
> > Can we please start creating changelogs?
> 
> Actually I like the current way we handle the release notes. As suggested by @KaiVolland please open an issue open for discussion if you miss anything.

Issue #1717, opened 2020-09-24
`pressedIconName` / `pressedIcon` property of toggleButton


annarieger:
if one of `iconName` or `icon` is set as prop on ToggleButton, one of`pressedIconName`or `pressedIcon` must be set as well. Otherwise the button turns invisible (`ant-btn:empty` class will be added) after the button was toggled.

Example:
```
<ToggleButton
  iconName="frown-o"
  // pressedIconName="smile-o" --> without this prop the button gets broken on toggle
  onToggle={()=>{}}
/>
```

Unraveler:
Hi @annarieger ,
You are correct, `ant` is setting this to hidden if no text or icon are given. It makes sense in general to provide both a pressed icon and a default icon, or maybe a text, making the button not always empty. But sometimes our code logic may required otherwise. For this reason, I am setting a new class in the react-geo-togglebutton style

```
// overrides the ant-btn:empty style, that is setting the button to hidden if no text or icon are given
  &:empty {
    visibility: visible;
  }
```
This will fix the issue

PR #1734, opened 2020-10-07
Bugfix/1717 pressed icon name pressed icon property of toggle button

Unraveler:
### Description:

if one of `iconName` or icon is set as prop on `ToggleButton`, one of `pressedIconName` or `pressedIcon` must be set as well. Otherwise the button turns invisible (`ant-btn:empty` class will be added) after the button was toggled.
Altough it makes sense in general to provide both a pressed icon and a default icon, or maybe a text, making the button not always empty, sometimes our code logic may required otherwise. For this reason, I am setting a new class in the `react-geo-togglebutton` style that will override the `ant-btn:empty`.

## CHECKLIST
Fixes Issue: #1717 
Examples added? No (No need to)
Tests added? Yes
Docs added? No (no need to)
Would a screenshot be helpful? No
Do you want to mention someone: @annarieger

### DIFF
--- a/src/Button/ToggleButton/ToggleButton.spec.tsx
+++ b/src/Button/ToggleButton/ToggleButton.spec.tsx
@@ -223,6 +223,6 @@ describe('<ToggleButton />', () => {
       pressedIconName: undefined,
       pressed: true
     });
-    expect(wrapper).not.toBeUndefined();
+    expect(wrapper.find('button')).toBeDefined();
   });
 });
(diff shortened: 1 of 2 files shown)


annarieger (approved):

src/Button/ToggleButton/ToggleButton.spec.tsx
```diff
@@ -215,4 +215,14 @@ describe('<ToggleButton />', () => {
...
   });
+
+  it('can be rendered if iconName is set and no text or icon is set with the property pressed set to true', () => {
+    const wrapper = TestUtil.mountComponent(ToggleButton);
+    wrapper.setProps({
+      iconName: 'some-icon-name',
+      pressedIconName: undefined,
+      pressed: true
+    });
+    expect(wrapper).not.toBeUndefined();
```
dnlkoch, line 226: I'm not sure if this will test the actual result. You may want to check if the expected html fragment is actually being rendered through the wrapper.

dnlkoch (approved):

Issue #1710, opened 2020-09-22
the pressed property of the ToggleButton is not respected


simonseyock:
The pressed property of the ToggleButton is not respected. It works by setting the `selectedName` of the togglegroup.

See https://codesandbox.io/s/react-geo-pressed-ol38p?file=/index.js for an example

simonseyock:
Maybe pressed could be calculated like this:
```
this.stat.selectedName === undefined ? child.props.pressed : this.state.selectedName === child.props.name
```
But then selectedName would need to get updated after and this would allow multiple buttons to be pressed at once.
The current behaviour is definitely easier, but if it should stay this way it would be nice if the documentation could be updated.

Unraveler:
I agree on updating the Docs, it makes sense to have the group control the button that is pressed, but this is not at all clear

PR #1735, opened 2020-10-07
Added notes to documentation on ToogleGroup and ToogleButton

Unraveler:
DOCUMENTATION

### Description:

`ToggleButton` has a property called `pressed` that is overridden by the property `selectedName` from `ToogleGroup`. This appears to cause some confusion on some users, so a note to the documentation has been added.

Fixes Issue? #1710 
Examples added? No (not needed)
Tests added? No (not needed)
Docs added? Yes
Would a screenshot be helpful? No

### DIFF
--- a/src/Button/ToggleButton/ToggleButton.tsx
+++ b/src/Button/ToggleButton/ToggleButton.tsx
@@ -26,7 +26,7 @@ interface DefaultProps {
   tooltipProps: AbstractTooltipProps;
   /**
    * The initial pressed state of the ToggleButton
-   * Note: If ToggleButton is set inside a ToggleGroup, this prop will be controlled by selectedName prop on
+   * Note: If a ToggleButton is inside a ToggleGroup, the pressed state will be controlled by the selectedName property of the
    * ToggleGroup. Setting in in Toggle button will not work.
    */
   pressed: boolean;
--- a/src/Button/ToggleButton/ToggleButton.tsx
+++ b/src/Button/ToggleButton/ToggleButton.tsx
@@ -26,8 +26,8 @@ interface DefaultProps {
   tooltipProps: AbstractTooltipProps;
   /**
    * The initial pressed state of the ToggleButton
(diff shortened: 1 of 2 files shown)


src/Button/ToggleButton/ToggleButton.tsx
```diff
@@ -26,6 +26,8 @@ interface DefaultProps {
   tooltipProps: AbstractTooltipProps;
   /**
    * The initial pressed state of the ToggleButton
+   * Note: If ToggleButton is set inside a ToggleGroup, this prop will be controlled by selectedName prop on
```
simonseyock, line 29:
```suggestion
   * Note: If a ToggleButton is inside a ToggleGroup, this property will be controlled by selectedName property of the
```

src/Button/ToggleButton/ToggleButton.tsx
```diff
@@ -26,6 +26,8 @@ interface DefaultProps {
...
+   * ToggleGroup. Setting in in Toggle button will not work.
```
simonseyock, line 30:
```suggestion
   * ToggleGroup and the will property will be ignored.
```

src/Button/ToggleGroup/ToggleGroup.tsx
```diff
@@ -30,6 +30,8 @@ interface BaseProps {
   /**
    * The value fo the `name` attribute of the children to select/press
    * initially.
+   * Note: This prop will have full control over the pressed prop on its children. Setting select/pressed on the
+   * children props directly, will have no effect.
```
simonseyock, line 34:
```suggestion
   * children props directly will have no effect.
```

simonseyock: I am not sure here :) but it should work without `,`

Unraveler: It's correct. Better without comma

dnlkoch:
As soon as the lint errors are fixed, this can be merged.

Issue #697, opened 2018-06-06
Documentation / Examples too large, breaks on some mobiles


weskamm:
https://terrestris.github.io/react-geo/docs/latest/index.html

loads 11,5 MB and triggers 222 requests on page load.
Apple iPad mini Safari browser breaks, as its too huge.
Besides, the layout / menu is not visible on mobile (android chrome).

Maybe we can introduce some kind of lazy loading / initial collpased state of all examples?

marcjansen:
Yes please, also I'd find it nice if it didn't directly scroll over to the Ujah-image. Can we not create many single pages for all the components?

marcjansen:
What should we do here?

dnlkoch:
We should give [pagePerSelection](https://react-styleguidist.js.org/docs/configuration.html#pagepersection) a try at least.

weskamm:
fixed by Unraveler, thanks!

PR #1736, opened 2020-10-07
added pagePerSection and sectionDepth to styleguide.config.js

Unraveler:
### Description:

Documentation website was taking too long to load, because it was loading all the examples on start. Added pagePerSection on `styleguide.config` to avoid this scenario.

Fixes Issue? #697 
Examples added? No need to
Tests added? No need to
Docs added? No need to
Would a screenshot be helpful? see below
Do you want to mention someone? @weskamm , @marcjansen , @dnlkoch 

Peek 2020-10-07 14-15

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -26,7 +26,10 @@ module.exports = {
     '**/*.d.ts'
   ],
   theme: {
-    sidebarWidth: 350
+    sidebarWidth: 350,
+    mq: {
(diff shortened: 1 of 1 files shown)


weskamm:
We had 247 requests on initial page load with 17MB traffic. With these changes, we are down to 18 requests and 4MB.
Thanks for tackling this!

PR #1917, opened 2021-01-05
Use center and zoom for zoomToExtent Button

hblitza:
### Description:

- Adds option to define the extent via _center_ and _zoom_
- Change _extent_ to be an optional prop, same as _center_ and _zoom_
- Adds notification if none of the necessary props are provided

Background: Since a fixed extent is mostly dependant on screen ratio / size, this offers a different way.

@terrestris/devs please review.

### DIFF
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.spec.tsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.spec.tsx
@@ -85,12 +85,14 @@ describe('<ZoomToExtentButton />', () => {
       setTimeout(resolve, 1200);
     });
 
-    expect.assertions(2);
+    expect.assertions(3);
     return promise.then(() => {
       const newExtent = map.getView().calculateExtent();
       const newCenter = getCenter(newExtent);
+      const newZoom = map.getView().getZoom();
       expect(newCenter).toEqual(mockExtentCenter);
       expect(containsExtent(newExtent, mockExtent)).toBe(true);
+      expect(newZoom).toEqual(mockZoom);
     });
   });
 });
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.spec.tsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.spec.tsx
@@ -13,6 +13,7 @@ describe('<ZoomToExtentButton />', () => {
   const mockGeometryCenter = [5000, 5000];
   const mockExtent = [0, 0, 10000, 10000];
   const mockExtentCenter = [5000, 5000];
+  const mockZoom = 7;
 
   beforeEach(() => {
     map = TestUtil.createMap();
@@ -70,5 +71,26 @@ describe('<ZoomToExtentButton />', () => {
       expect(containsExtent(newExtent, mockExtent)).toBe(true);
     });
 
+  });
+
+  it('zooms to extent when clicked providing center and zoom', () => {
+    const wrapper = TestUtil.mountComponent(ZoomToExtentButton, {
+      map,
+      center: mockExtentCenter,
+      zoom: mockZoom
+    });
+    wrapper.instance().onClick();
+
+    const promise = new Promise(resolve => {
+      setTimeout(resolve, 1200);
+    });
+
+    expect.assertions(2);
+    return promise.then(() => {
+      const newExtent = map.getView().calculateExtent();
+      const newCenter = getCenter(newExtent);
+      expect(newCenter).toEqual(mockExtentCenter);
+      expect(containsExtent(newExtent, mockExtent)).toBe(true);
+    });
   });
 });
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
@@ -131,7 +131,7 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
     if (extent) {
       view.fit(extent, finalFitOptions);
     }
-    else if (!extent && (center && zoom)) {
+    else if (center && zoom) {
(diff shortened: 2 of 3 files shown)


src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -28,6 +31,20 @@ interface DefaultProps {
...
+  extent: number[] | OlSimpleGeometry;
+  /**
+   * The center `[x,y]` in the maps coordinate system or an
+   * instance of ol.coordinate that the map should zoom to if no extent is given.
+   */
+  center: OlCoordinate;
+  /**
+   *  The zoom level 'x' the map should zoom to  if no extent is given.
+   */
+  zoom: number;
```
annarieger, line 47: Please mark all three parameters extent, center and zoom as optional (e.g. `zoom?: number;`)

annarieger (approved):
Nice addition!

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -28,6 +31,20 @@ interface DefaultProps {
...
+  /**
+   * The extent `[minx, miny, maxx, maxy]` in the maps coordinate system or an
+   * instance of ol.geom.SimpleGeometry that the map should zoom to.
+   */
+  extent?: number[] | OlSimpleGeometry;
+  /**
+   * The center `[x,y]` in the maps coordinate system or an
+   * instance of ol.coordinate that the map should zoom to if no extent is given.
+   */
+  center?: OlCoordinate;
```
weskamm, line 43: should read
```
center?: number[] | OlCoordinate;
```
according to the docs?

src/Button/ZoomToExtentButton/ZoomToExtentButton.spec.tsx
```diff
@@ -71,4 +72,25 @@ describe('<ZoomToExtentButton />', () => {
...
+
+    const promise = new Promise(resolve => {
+      setTimeout(resolve, 1200);
+    });
+
+    expect.assertions(2);
+    return promise.then(() => {
+      const newExtent = map.getView().calculateExtent();
+      const newCenter = getCenter(newExtent);
+      expect(newCenter).toEqual(mockExtentCenter);
```
weskamm, line 92: please add a check for the zoom too

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
hblitza: right thanks!

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -93,6 +110,9 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
     if (!view) { // no view, no zooming
       return;
     }
+    if (!extent && (!center && !zoom)) {
```
weskamm, line 113: should read
```
!extent && (!center || !zoom)
```
as using only zoom or only center should also fail / raise a message

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -93,6 +110,9 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
...
+      logger.warn('Provide either an extent or a center and a zoom.');
```
weskamm, line 114: Can be changed to an error message, as the button will not work. And its useful to see which component the message comes from, e.g.
```
logger.error('zoomToExtentButton: You need to provide either an extent or a center and a zoom.');
```

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -93,6 +110,9 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
...
+    }
```
weskamm, line 115: please add a return statement in the if-block, as the code shall not continue without valid parameters

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
weskamm: or you keep that line and change the behaviour below to zoom only when zoom is given and center only when center is given.

weskamm (approved):

hblitza:
Thanks for the review!

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -104,7 +125,17 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
       ...fitOptions
     };
 
-    view.fit(extent, finalFitOptions);
+    if (extent && (center && zoom)) {
```
marcjansen, line 128: How will this react to a valid zoom of `0`?

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -104,7 +125,17 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
...
+      logger.warn('Provide either an extent or a center and a zoom.' +
+      'If both are provided the extent will be used.');
+    }
+    if (extent) {
+      view.fit(extent, finalFitOptions);
+    }
+    else if (center && zoom) {
```
marcjansen, line 135: Same question here wrt zoom `0`.

marcjansen (commented):
Nice and thanks. Some questions wrt zoomlevel `0` which might be handled incorrectly.

PR #1919, opened 2021-01-06
zoomToExtent improve property checking

hblitza:
## BUGFIX
Improve property checking for  `zoom`. A valid `zoom` of `0` must be accepted. Thanks @marcjansen @LukasLohoff 

@terrestris/devs please review!

### DIFF
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
@@ -127,8 +127,8 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
     };
 
     if (extent && (center && _isFinite(zoom))) {
-      logger.warn('Provide either an extent or a center and a zoom.' +
-      'If both are provided the extent will be used.');
+      logger.warn('zoomToExtentButton: Both extent and center / zoom are provided. ' +
+      'Extent will be used in favor of center / zoom');
     }
     if (extent) {
       view.fit(extent, finalFitOptions);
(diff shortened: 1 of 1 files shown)


src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -10,6 +10,8 @@ import { CSS_PREFIX } from '../../constants';
 
 import logger from '@terrestris/base-util/dist/Logger';
 
+const _isFinite = require('lodash/isFinite');
```
LukasLohoff, line 13:
```suggestion
import _isFinite from 'lodash/isFinite';
```

LukasLohoff: for consistency I'd use import here

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -125,14 +127,14 @@ class ZoomToExtentButton extends React.Component<ZoomToExtentButtonProps> {
       ...fitOptions
     };
 
-    if (extent && (center && zoom)) {
+    if (extent && (center && _isFinite(zoom))) {
       logger.warn('Provide either an extent or a center and a zoom.' +
```
LukasLohoff, line 131: I think the warning message is still somewhat unclear. At this point we already know that both extent and center/zoom are provided

So something like `Both extent and center / zoom are provided. Extent will be used in favor of center / zoom`

LukasLohoff (approved):

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
hblitza: good point thanks!

PR #1923, opened 2021-01-07
Add error message and custom handler for broken legends

weskamm:
### Description:
This adds the option to specify a custom error handler function, when the loading of the legend image failed.
Besides, an error message can be given that will be displayed in case of an load error. 
This will effectively remove the default broken image of the browser.

### DIFF
--- a/src/Legend/Legend.tsx
+++ b/src/Legend/Legend.tsx
@@ -22,6 +22,15 @@ export interface BaseProps {
    * be transformed to "&HEIGHT=400" an added to the GetLegendGraphic request.
    */
   extraParams?: any;
+  /**
+   * Optional method that should be called when the image retrieval errors
+   */
+  onError?: () => void;
+  /**
+   * Optional error message that should be displayed when image retrieval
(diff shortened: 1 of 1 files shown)


src/Legend/Legend.tsx
```diff
@@ -134,11 +155,19 @@ export class Legend extends React.Component<LegendProps, LegendState> {
...
         {...passThroughProps}
       >
-        <img
-          src={this.state.legendUrl}
-          alt={alt}
-          onError={this.onError.bind(this)}
-        />
+        {this.state.legendError ?
+          <div
+            className='legend-load-error'
```
annarieger, line 160: where are the rules for this class defined?

if no rules are specified, please remove the class.

annarieger (approved):

src/Legend/Legend.tsx
weskamm: its a bonus for external applications so they can modify the error div to their likings

hblitza:
Works for me, thanks!

PR #1924, opened 2021-01-07
Fix deploy path and version variable

dnlkoch:
### Description:

This should fix the target path in the `gh-pages` branch (`docs/latest` instead of `docs/latest/styleguide`) and the version number.

Please note @terrestris/devs.


hblitza:
Looks good, please try!

weskamm:
As the docs work again, this is good to go. Thanks!

PR #1929, opened 2021-01-11
Backport digitize fixes

simonseyock:
### Description:
Backport of DigitizeButton related commits.


src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.spec.tsx
```diff
@@ -2,18 +2,20 @@
...
 import moment from 'moment';
-import TimeLayerSliderPanel from '../TimeLayerSliderPanel/TimeLayerSliderPanel';
+import TimeLayerSliderPanel from './TimeLayerSliderPanel';
 
 describe('<TimeLayerSliderPanel />', () => {
   it('is defined', () => {
     expect(TimeLayerSliderPanel).not.toBeUndefined();
   });
 
-  it('can be rendered', () => {
```
weskamm, line 12: This is completely unrelated and running on other branches, so please try to restore this test.

weskamm:
Not sure what changes in detail, but looks ok to me

src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.spec.tsx
simonseyock: This test fails with a very weird error that I was not able to fix in the time I had:
```
FAIL src/Panel/TimeLayerSliderPanel/TimeLayerSliderPanel.spec.tsx

  ● <TimeLayerSliderPanel /> › can be rendered

    Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

    Check the render method of `TimeLayerSliderPanel`.

      30 |    */
      31 |   static mountComponent = (Component: any, props?: any, options?: MountRendererProps): Wrapper => {
    > 32 |     return mount(<Component {...props}/>, options);
         |            ^
      33 |   }
      34 | 
      35 |   /**
```

This is especially weird as the previous test explicitly tests that `TimeLayerSliderPanel` is not undefined. I suppose this has to do with babel or typescript.

simonseyock: This is fixed by cherry-picking b4b6a7628231c4332edc3dfde7215188f9fdd4b7 thanks @dkoch

weskamm: and you are sure its unrelated to the changed import?

simonseyock: Somehow I lost the uncommenting of the code while cherry-picking. I dropped the commit that removes the test fully now.

dnlkoch (approved):

PR #1930, opened 2021-01-11
check if interaction exists in onToggle

simonseyock:
### Description:
check if interaction exists before calling setActive


weskamm (approved):
Makes sense! Thanks!

PR #1935, opened 2021-01-14
Allow headless usage of nominatim and wfs search

weskamm:
### Description:
This allows one to use the Nominatim and WFS search in a headless way, so that no rendering of input fields or results happen.

This becomes useful when implementing your own UI or when combining multiple searches.

### DIFF
--- a/src/Field/NominatimSearch/NominatimSearch.tsx
+++ b/src/Field/NominatimSearch/NominatimSearch.tsx
@@ -68,6 +68,24 @@ interface DefaultProps {
    * returned by nominatim.
    */
   onSelect: (item: any, olMap: OlMap) => void;
+  /**
+   * Indicate if we should render the input and results. When setting to false,
+   * you need to handle user input and result yourself
+   */
+  visible?: boolean;
+  /**
+   * The searchTerm may be given as prop. This is useful when setting
+   * `visible` to `false`.
+   */
+  searchTerm?: string;
+  /**
+   * A callback function which gets called with the successfully fetched data.
+   */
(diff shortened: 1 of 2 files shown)


src/Field/NominatimSearch/NominatimSearch.tsx
```diff
@@ -68,6 +70,24 @@ interface DefaultProps {
    * returned by nominatim.
    */
   onSelect: (item: any, olMap: OlMap) => void;
+  /**
+   * Indicate if we should render the input and results. When setting to false,
+   * you need to handle user input and result yourself
+   */
+  doRender?: boolean;
```
simonseyock, line 77: maybe call it `visible`? This maybe is a little bit more standard,

src/Field/NominatimSearch/NominatimSearch.tsx
```diff
@@ -242,6 +274,10 @@ export class NominatimSearch extends React.Component<NominatimSearchProps, Nomin
   onFetchSuccess(response: any) {
     this.setState({
       dataSource: response
+    }, () => {
+      if (isFunction(this.props.onFetchSuccess)) {
```
simonseyock, line 278: Because of the typing it should be enough to just test for
````
if(this.props.onFetchSuccess) {
```

dnlkoch (commented):
Basically this is fine for me, but I would prefer to have the search logic to a reusable standalone class (e.g. in the `base-util`).

weskamm:
Thanks for reviews. Yes, it makes sense to duplicate the logic to base-util when time allows

PR #1953, opened 2021-01-28
Fix handling visibility of layer groups in tree

weskamm:
### Description:
This fixes some issues with layergroup visibility:
  * Before this change, one could not create a tree with an invisible layergroup in it, as the visibility handling for the containing layers was broken (could not be changed)
  * When a child layer changed its visibility, the parenting folder was not correctly made visible and in turn, the layer did not become visible. This has changed now. Setting a layer visible will turn on its parenting layer groups in order to really "see" the layer in the map as visible

### DIFF
--- a/src/LayerTree/LayerTree.spec.tsx
+++ b/src/LayerTree/LayerTree.spec.tsx
@@ -14,9 +14,11 @@ import Logger from '@terrestris/base-util/dist/Logger';
 
 describe('<LayerTree />', () => {
   let layerGroup;
+  let layerSubGroup;
   let map;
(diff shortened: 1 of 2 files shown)


dnlkoch (approved):

KaiVolland (approved):
LGTM.

Would be nice if you could fix the warnings, too. Not blocking.

annarieger (approved):

weskamm:
Thanks for reviews. @KaiVolland if you mean the deprecation warning, i fear that one will need a complete rewrite of this class.

PR #1959, opened 2021-01-29
Openlayers types

simonseyock:
## FEATURE | BUGFIX

This PR introduces the `@hanreev/ol-types` package and fixes all openlayers typing errors.


simonseyock:
moved this types to 'devDependencies'. Is this ok?
```
    "@types/lodash": "^4.14.155",
    "@types/moment": "^2.13.0",
    "@types/node": "^14.0.13",
    "@types/react-dom": "^17.0.0",
    "@types/react-rnd": "^8.0.0",
 ```

One thing that should get checked, are the attributions. But this is unrelated to this PR, see https://github.com/terrestris/react-geo/issues/1962

I used the public function `getUid` instead of the private properties `ol_uid`.
 
For https://github.com/terrestris/react-geo/pull/1959/files#diff-daa97fd8a5ec8ae108a65464bc59cfe464fcb0b55280ddb5f647ac2a1aa586f4R180 I switched to use the crs code as the value of the select. Previously the code would be set to the key property and this would be checked. But there is no reason to not use the code as the value in my opinion.

Similarly I used value in https://github.com/terrestris/react-geo/pull/1959/files#diff-b6862592b05897d382b157f6abb67d22575c08b8b48972548d5d3e9ced4fc182R299

I fixed the types of the `timeLayerAware` function. There is now an correctly typed Option.

For the type ArrayTwoOrMore I introduced a type guard `isArrayTwoOrMore` that let you easily verify that an array is indeed of the type ArrayTwoOrMore.

Another new type guard is `isWmsLayer` which verifies that a generic `Layer` is a `BaseTileLayer|BaseImageLayer`.

**Comment by simonseyock** [comment]
I reverted the changes related to the selects.

File: src/Button/DigitizeButton/DigitizeButton.tsx [hunk 1]
```diff
@@ -12,15 +12,19 @@ import OlStyleStroke from 'ol/style/Stroke';
...
 import OlInteractionSelect from 'ol/interaction/Select';
 import OlInteractionModify from 'ol/interaction/Modify';
 import OlInteractionTranslate from 'ol/interaction/Translate';
 import OlFeature from 'ol/Feature';
 import { never, singleClick } from 'ol/events/condition';
+import OlGeometry from 'ol/geom/Geometry';
+import { Select } from 'ol/interaction';
+import Draw from 'ol/interaction/Draw';
+import Modify from 'ol/interaction/Modify';
+import Translate from 'ol/interaction/Translate';
```
KaiVolland, line 25: Please rename the new imports like this. So we are in sync with the others.

```suggestion
import { Select as OlInteractionSelect } from 'ol/interaction';
import OlInteractionDraw from 'ol/interaction/Draw';
import OlInteractionModify from 'ol/interaction/Modify';
import OlInteractionTranslate from 'ol/interaction/Translate';
```

Maybe you can change the import of the event conditions to `import EventConditions from 'ol/events/condition'` too. So the execute code would look like `EventConditions.never`

src/Button/GeoLocationButton/GeoLocationButton.tsx
```diff
@@ -9,6 +9,8 @@ import OlLayerVector from 'ol/layer/Vector';
 import OlSourceVector from 'ol/source/Vector';
 import OlStyleStyle from 'ol/style/Style';
 import OlStyleIcon from 'ol/style/Icon';
+import Geometry from 'ol/geom/Geometry';
+import GeometryLayout from 'ol/geom/GeometryLayout';
```
KaiVolland, line 13: See first comment.

src/Button/MeasureButton/MeasureButton.tsx
```diff
@@ -289,15 +295,15 @@ class MeasureButton extends React.Component<MeasureButtonProps> {
       this._drawInteraction.setActive(pressed);
 
       this._eventKeys.drawstart = this._drawInteraction.on(
-        'drawstart', this.onDrawStart, this
+        'drawstart', e => this.onDrawStart(e)
```
KaiVolland, line 298: Can you explain, why you changed this?

src/Button/MeasureButton/MeasureButton.tsx
```diff
@@ -365,10 +371,9 @@ class MeasureButton extends React.Component<MeasureButtonProps> {
     } = this.props;
 
     const maxPoints = measureType === 'angle' ? 2 : undefined;
-    const drawType = measureType === 'polygon' ? 'MultiPolygon' : 'MultiLineString';
+    const drawType = measureType === 'polygon' ? GeometryType.MULTI_POLYGON : GeometryType.MULTI_LINE_STRING;
```
KaiVolland, line 374: :+1:

src/Container/AddWmsPanel/AddWmsLayerEntry/AddWmsLayerEntry.tsx
```diff
@@ -49,6 +52,7 @@ export class AddWmsLayerEntry extends React.Component<AddWmsLayerEntryProps, Add
    */
   constructor(props: AddWmsLayerEntryProps) {
     super(props);
+    // TODO: getAttributions is not @api and returns a function in v6.5
```
KaiVolland, line 55: This needs to be adpated/fixed before we can merge this.

src/Field/NominatimSearch/NominatimSearch.tsx
```diff
@@ -105,9 +108,27 @@ interface BaseProps {
   onClear?: () => void;
 }
 
+// See https://nominatim.org/release-docs/develop/api/Output/ for some more information
+type NominatimPlace = {
```
KaiVolland, line 112: Please move this to the top and maybe add an `export`.

src/LayerSwitcher/LayerSwitcher.tsx
```diff
@@ -132,7 +128,7 @@ export class LayerSwitcher extends React.Component<LayerSwitcherProps, LayerSwit
    */
   cloneLayer = (layer: OlLayerTile | OlLayerGroup) => {
     let layerClone: OlLayerTile | OlLayerGroup;
-    if (layer.getLayers) {
+    if (layer instanceof OlLayerGroup) {
```
KaiVolland, line 131: We had some difficulties with `instanceof` checks. Maybe you leave it as is. Or introduce `isLayerGroup()` to Openlayers. :wink:

src/LayerTree/LayerTree.tsx
```diff
@@ -75,8 +76,8 @@ export interface BaseProps {
 
 interface LayerTreeState {
   layerGroup: OlLayerGroup;
-  layerGroupRevision: null;
-  treeNodes: LayerTreeNode[];
+  layerGroupRevision?: number;
+  treeNodes: ReactElement[];
```
KaiVolland, line 80: This seems to be more unprecise the the previous typing.

src/Util/isWmsLayer.ts
```diff
@@ -0,0 +1,11 @@
+import Layer from 'ol/layer/Layer';
+import Source from 'ol/source/Source';
+import BaseTileLayer from 'ol/layer/BaseTile';
+import BaseImageLayer from 'ol/layer/BaseImage';
+import OlSourceImageWMS from 'ol/source/ImageWMS';
+import OlSourceTileWMS from 'ol/source/TileWMS';
+
+export function isWmsLayer(layer: Layer<Source>): layer is (BaseTileLayer|BaseImageLayer) {
+  const source = layer.getSource();
+  return source instanceof OlSourceImageWMS || source instanceof OlSourceTileWMS;
```
KaiVolland, line 10: See above.

src/Util/ArrayTwoOrMore.ts
```diff
@@ -0,0 +1,5 @@
+export type ArrayTwoOrMore<T> = [T, T] & T[];
```
KaiVolland, line 1: Is this util ever used? For me this looks something for the `base-util` package.

src/Util/isWmsLayer.ts
```diff
@@ -0,0 +1,11 @@
+import Layer from 'ol/layer/Layer';
+import Source from 'ol/source/Source';
+import BaseTileLayer from 'ol/layer/BaseTile';
+import BaseImageLayer from 'ol/layer/BaseImage';
+import OlSourceImageWMS from 'ol/source/ImageWMS';
+import OlSourceTileWMS from 'ol/source/TileWMS';
+
+export function isWmsLayer(layer: Layer<Source>): layer is (BaseTileLayer|BaseImageLayer) {
```
KaiVolland, line 8: Is this util ever used? For me this looks something for the `ol-util` package.

KaiVolland (commented):
Nice one. Ol types would be very cool if they are up to date.
Please have a look the import statements to have the `Ol` prefix.

Even if the changes are mostly about typing i think they introduce api changes, so we have to release a major version (which is not negative).

src/Button/DigitizeButton/DigitizeButton.tsx
simonseyock: Done

src/Button/MeasureButton/MeasureButton.tsx
simonseyock: in openlayers v5 the opt_this parameter on `on` was removed

src/Container/AddWmsPanel/AddWmsLayerEntry/AddWmsLayerEntry.tsx
simonseyock: this actually has nothing todo with this commit, but was a problem that existed before. Maybe it is only documented is badly. I opened this issue for it: https://github.com/terrestris/react-geo/issues/1962

src/LayerSwitcher/LayerSwitcher.tsx
simonseyock: I introduced a type guard `isLayerGroup` to still benefit from the typings

src/Util/isWmsLayer.ts
simonseyock: This is purely for typing. What were the problems? The only thing I can imagine is different versions of the library in the same build. If we work with types this would be the nicer way.

src/Util/ArrayTwoOrMore.ts
simonseyock: Yes it is used by the LayerSwitcher.

src/Util/isWmsLayer.ts
simonseyock: Yes it needs to be used because the type documentation in openlayers was imprecise in regard of the source types. A PR into openlayers was merged to fix this. But I recognized that I could improve the resulting types of this guard.

src/Util/isWmsLayer.ts
KaiVolland: I can't remember but the two different versions is the most likely issue. Not sure if you can allways avoid it 100% but as `ol` is just a peerdependcy we should give it a try.

src/Util/isWmsLayer.ts
KaiVolland: I still think it would fit better in the `ol-util`. If you want to merge it like this please open an issue to move it. Same goes for the `ArrayTwoOrMore` util.

src/LayerTree/LayerTree.tsx
simonseyock: I agree, but somehow the typecheck was failing here. I don't exactly see why it was passing before.

simonseyock: The problem is that typescript says that `return (<LayerTreeNode ... >)` is not assignable to the type `LayerTreeNode`. Do you know why this could be the case?

src/Util/isWmsLayer.ts
simonseyock: Yes I would say this would be the way to go because `ol-util` is not yet moved to using the ol types

simonseyock: I looked at `ol-util` and it is not written in typescript so it does not make sense to add these types there

simonseyock:
One thing about actuality of the types: They should change only in details on minor version changes of openlayers. The hanreev types base for now on ol `v6.4` but I do not think this is something we need to worry about as we are only using ol `v6.5` for now.

On the next major version jump we would need to look at it again, but I am working on type improvements in the openlayers jsdoc types that will make the auto generated types usable in our projects. The hanreev types are way better, as they provide more details, for example they type the events that will be passed to the listeners.

src/LayerTree/LayerTree.tsx
KaiVolland: Hmm could it be related to antd? And why did it work before?

src/HigherOrderComponent/TimeLayerAware/TimeLayerAware.spec.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import moment from 'moment';
 
-import TileLayer from 'ol/layer/Tile.js';
-import TileWMS from 'ol/source/TileWMS';
+import OlTileLayer from 'ol/layer/Tile.js';
```
ahennr, line 4: `js` extension requried?

ahennr (approved):
Nice work @simonseyock .

src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
```diff
@@ -1,31 +1,25 @@
 import * as React from 'react';
 
+import _isFinite from 'lodash/isFinite';
+
 import OlMap from 'ol/Map';
 import OlSimpleGeometry from 'ol/geom/SimpleGeometry';
 import { Coordinate as OlCoordinate } from 'ol/coordinate';
 import { easeOut } from 'ol/easing';
+import { Extent } from 'ol/extent';
+import { FitOptions } from 'ol/View';
```
annarieger, line 10: Please change it to
```
import { Extent as OlExtent} from 'ol/Extent;
import { FitOptions as OlFitOptions } from 'ol/View';
```

src/Field/NominatimSearch/NominatimSearch.tsx
```diff
@@ -2,19 +2,40 @@ import * as React from 'react';
...
+  place_id: number;
+  osm_type: string;
+  osm_id: number;
+  boundingbox: string[];
+  display_name: string;
+  category: string;
+  type: string;
+  importance: number;
+  icon?: string;
+  adress?: any;
```
annarieger, line 32: should be `address`

src/Field/NominatimSearch/NominatimSearch.tsx
```diff
@@ -2,19 +2,40 @@ import * as React from 'react';
...
+import { OptionProps } from 'antd/lib/select';
+import { OptionData } from 'rc-select/lib/interface';
 
 import Logger from '@terrestris/base-util/dist/Logger';
 import UrlUtil from '@terrestris/base-util/dist/UrlUtil/UrlUtil';
 
 import OlMap from 'ol/Map';
-
 import { transformExtent } from 'ol/proj';
+import { Extent } from 'ol/extent';
```
annarieger, line 13: Please change to 
```
import { Extent as OlExtent } from 'ol/extent';
```

annarieger (commented):
Nice one, thanks for this.

I've added some notes which you maybe want to address.

dnlkoch (commented):
Very nice @simonseyock, I just addressed some minor issues regarding our current naming conventions and a note about moving code to separate packages.

src/Field/NominatimSearch/NominatimSearch.tsx
simonseyock: nice catch!

src/Field/WfsSearchInput/WfsSearchInput.tsx
```diff
@@ -10,7 +10,7 @@ import CloseCircleOutlined from '@ant-design/icons/CloseCircleOutlined';
 
 import OlMap from 'ol/Map';
 
-const debounce = require('lodash/debounce');
+import debounce from 'lodash/debounce';
```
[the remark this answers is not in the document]
simonseyock: Yay bonus points XD

src/Util/typeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (layer instanceof OlLayer) {
+    const source = layer.getSource();
+    return source instanceof OlImageWMS || source instanceof OlTileWMS;
+  }
+  return false;
+}
+
+export function isImageOrTileLayer(layer: OlBaseLayer): layer is ImageLayer | TileLayer {
+  return layer instanceof ImageLayer || layer instanceof TileLayer;
+}
```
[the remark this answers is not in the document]
simonseyock: `ol-util` is not written in typescript so this type guards don't make sense there

src/Util/typeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+import OlImageWMS from 'ol/source/ImageWMS';
+import OlTileWMS from 'ol/source/TileWMS';
+import ImageLayer from 'ol/layer/Image';
+import TileLayer from 'ol/layer/Tile';
+
+export type ArrayTwoOrMore<T> = [T, T] & T[];
+
+export function isArrayTwoOrMore<T>(value: T[]): value is ArrayTwoOrMore<T> {
+  return value?.length > 1;
+}
```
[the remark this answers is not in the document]
simonseyock: `ol-util` is not written in typescript so this type guards don't make sense there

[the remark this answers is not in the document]
simonseyock: `ArrayTwoOrMore` is a type that is used at several points in `react-geo`. I included this guard here because it is nice to have in a project that uses react-geo.

src/LayerTree/LayerTree.tsx
simonseyock: I tried to revert the types to the initial state, but the typecheck still fails. Also I reverted typescript back to `v4.0.3` which also did not help.

src/Util/typeUtils.ts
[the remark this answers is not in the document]
dnlkoch: The `base-util` is actually written in typescript.

src/Util/typeUtils.ts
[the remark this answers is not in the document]
dnlkoch: Correct, missed that.

src/Util/isWmsLayer.ts
dnlkoch: We had problems with `instanceof` checks while having `react-geo` as linked dependency in projects due to fact that two versions of ol would exist in the scope then. But we're able to fix this issue with an `alias` for ol. So I think this change is fine.

src/Util/typeUtils.ts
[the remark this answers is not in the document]
simonseyock: i was getting confused with all the util libs :laughing:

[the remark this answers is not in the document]
simonseyock: I opened https://github.com/terrestris/base-util/pull/381

KaiVolland (approved):
Thanks for the productive discussion and your effort in this. Feel free to merge once the ci is happy.

PR #2074, opened 2021-03-15
Omit onSelect as the typing differs from the base component

dnlkoch:
### Description:

This omits the `onSelect` type from the base component as the type differs from the one defined in our component.


hwbllmnn (approved):

dnlkoch:
Thanks for the review!

PR #2108, opened 2021-03-26
Migrate userChip tests to react-testing-library

LukasLohoff:
### Description:

Migrates `UserChip` tests.

@terrestris/devs Please review

### DIFF
--- a/src/UserChip/UserChip.spec.tsx
+++ b/src/UserChip/UserChip.spec.tsx
@@ -18,7 +18,7 @@ describe('<UserChip />', () => {
 
   it('determines initials from given user name', () => {
     render(<UserChip userName="Shinji Kagawa" />);
-    const chip = screen.getByText('SK').parentElement;
+    const chip = screen.getByText('SK');
     expect(chip).toBeVisible();
   });
 
--- a/src/UserChip/UserChip.spec.tsx
+++ b/src/UserChip/UserChip.spec.tsx
@@ -1,57 +1,62 @@
 import testImage from  '../../assets/user.png';
-import TestUtil from '../Util/TestUtil';
 
 import UserChip from './UserChip';
+import { render, screen } from '@testing-library/react';
+import * as React from 'react';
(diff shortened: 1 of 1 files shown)


src/UserChip/UserChip.spec.tsx
```diff
@@ -1,57 +1,62 @@
 import testImage from  '../../assets/user.png';
-import TestUtil from '../Util/TestUtil';
 
 import UserChip from './UserChip';
+import { render, screen } from '@testing-library/react';
+import * as React from 'react';
+import '@testing-library/jest-dom';
```
simonseyock, line 6:
```suggestion
```

src/UserChip/UserChip.spec.tsx
```diff
@@ -1,57 +1,62 @@
...
+    const { container } = render(<UserChip />);
+    expect(container).toBeVisible();
   });
 
   it('determines initials from given user name', () => {
-    wrapper.setProps({userName: 'Shinji Kagawa'});
-    const test = wrapper.instance().getInitials();
-    expect(test).toBe('SK');
+    render(<UserChip userName="Shinji Kagawa" />);
+    const chip = screen.getByText('SK').parentElement;
```
simonseyock, line 22:
```suggestion
    const chip = screen.getByText('SK');
```
wouldn't that be enough?

simonseyock (commented):
Looks great, only 2 minor remarks.

src/UserChip/UserChip.spec.tsx
LukasLohoff: in this case, yes

simonseyock:
The new eslint plugin also has some remarks :dizzy:

LukasLohoff:
> The new eslint plugin also has some remarks

Fixed that. The rule `prefer-in-document` is kind of annoying though. Should be a warning IMHO, not an error.

LukasLohoff:
Thanks for the review. Merging this now.

PR #2149, opened 2021-04-16
Update antd to 4.15.1

dnlkoch:
### Description:

This updates antd to it's latest version.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -21,7 +21,7 @@ module.exports = {
     '<rootDir>/dist/'
   ],
   transformIgnorePatterns: [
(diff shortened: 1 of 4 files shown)


marcjansen:
Can you quickly elaborate why the changes to the tests were needed? It looks as if sth. for user might change (number of nodes in tree, e.g.). I might as well be on the wrong track.

dnlkoch:
To be honest, I'm not entirely sure, why there is a mismatch, but I assume, the class `ant-tree-node` is now being applied to other elements (the folders) of the tree as well. 

I checked the rendered correspondence in the styleguide and couldn't observe any changes regarding the number of nodes.

PR #2150, opened 2021-04-16
Remove props draggable and checkable

dnlkoch:
### Description:

Both props are defined in the `TreeProps` and can safely be removed from our own definition.


LukasLohoff (approved):

dnlkoch:
Thanks for the review!

PR #2172, opened 2021-04-26
Fix selectedName state value of ToggleButton

annarieger:
### Description:
Update state value for `selectedName` key if it's prop value was changed as well.


KaiVolland (approved):
Nice fix. Please merge.

PR #2178, opened 2021-04-28
Fix ZoomToExtent projection issue

hblitza:
### Description:
`ZoomToExtent` Button needs to project the props `center` or `extent` for the current view's projection before applying it.  
If not, this leads to wrong zoom-to occurences when the projection has changed in an application.

Please have a look @terrestris/devs

### DIFF
--- a/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
+++ b/src/Button/ZoomToExtentButton/ZoomToExtentButton.tsx
@@ -8,6 +8,7 @@ import { Coordinate as OlCoordinate } from 'ol/coordinate';
 import { easeOut } from 'ol/easing';
 import { Extent as OlExtent } from 'ol/extent';
 import { FitOptions as OlViewFitOptions } from 'ol/View';
+import { transform, transformExtent, get as getProjection } from 'ol/proj';
 
 import SimpleButton, { SimpleButtonProps } from '../SimpleButton/SimpleButton';
 import { CSS_PREFIX } from '../../constants';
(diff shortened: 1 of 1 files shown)


marcjansen:
Not sure I get the point, but shouldn't the application simply change the props `extent` or `center` when it knows that the projection has changed? Isn't it theresponibility of the app to keep these things in sync? Again, I might be misunderstanding things entirely.

hblitza:
Okay I see. Will have a look thanks.

weskamm:
But this enforces users of react-geo to use a state container like redux, which may not always be the case. Or am i wrong @marcjansen ?

dnlkoch:
Yes, I agree with @marcjansen. Components should be as dump as possible and even more important: They should never manipulate their props!

> But this enforces users of react-geo to use a state container like redux, which may not always be the case. Or am i wrong @marcjansen ?

Can you explain this? It don't think that it's needed, but I may get it wrong.

weskamm:
> Can you explain this? It don't think that it's needed, but I may get it wrong.

Well the sentence "shouldn't the application simply change the props" assumes that there is a state container / global store / redux / call it whatever you want. And i thought that this was unlikely, but optional when building an application with react-geo

hblitza:
Thanks for review, will close this.

See https://github.com/terrestris/react-geo-baseclient/pull/792 for a new solution which does not affect `react-geo`.

dnlkoch:
> Well the sentence "shouldn't the application simply change the props" assumes that there is a state container / global store / redux / call it whatever you want. And i thought that this was unlikely, but optional when building an application with react-geo

Why do you need a state manger to change props? This can be done with the basic react data flow, e.g. (pseudocode):

```js
AppFunctionComponent = () => {
  const [projection, setProjection] = useState('EPSG:4326');

  useEffect(() => {
    setProjection('EPSG:1909');
  }, [myChangeCondition])

  return (
    <ZoomToExtentButton
      projection={projection}
    />
  );
};
```

And yes, you're absolutely right. We should not enforce the users of react-geo to make use of an external state manager.

weskamm:
> Why do you need a state manger to change props? 
Because a component can only read props, and not change them. Do you mean state instead? Or am i missing something here?

PR #2262, opened 2021-06-11
fix: evaluate style function inside of style function

simonseyock:
### Description:
this evaluates style function inside the `digitizeStyleFunction` and also renames the `getXyStyleFunction` functions to `xyStyleFunction` because they are used as style functions (and do not return a style function).


hwbllmnn (approved):

hwbllmnn:
Thanks!

PR #2299, opened 2021-07-02
Split DigitizeButton

### LINKED ISSUE #475: Split Digitize Button

simonseyock:
### Description:

This PR extracts the draw parts from DigitizeButton into a seperate component as described in #475.

Major differences to the DigitizeButton:

1. It accepts a `layer` option instead of `layerName` option.
2. The `drawStyle` is no longer applied to the layer. To change the style of the layer use:
  ```
  const digitizeLayer = DigitizeUtil.getDigitizeLayer(map);
  digitizeLayer.setStyle(...);
  ```
3. It will always use an own interaction and will not share it with other buttons of the same type. (It might happen that the `DigitizeButton` tries to use same interaction - so it should be avoided to use a `DigitizeButton` for drawing if you intend to use a `DrawButton`).

This PR includes useful helper functions to trigger events on the open layers map.


simonseyock:
I was able to extract the DrawButton. And it works:

image

simonseyock:
Major differences to the DigitizeButton:

1. It accepts a `layer` option instead of `layerName` option.
2. The `drawStyle` is no longer applied to the layer. To change the style of the layer use:
  ```
  const digitizeLayer = DigitizeUtil.getDigitizeLayer(map);
  digitizeLayer.setStyle(...);
  ```

This PR includes useful helper functions to trigger events on the open layers map.

src/Button/DrawButton/DrawButton.spec.tsx
```diff
@@ -0,0 +1,324 @@
...
+import MapUtil from '@terrestris/ol-util/dist/MapUtil/MapUtil';
+
+import OlView from 'ol/View';
+import OlMap from 'ol/Map';
+import MapComponent from '../../Map/MapComponent/MapComponent';
+import { clickMap, doubleClickMap } from '../../Util/rtlTestUtils';
+import LineString from 'ol/geom/LineString';
+import Polygon from 'ol/geom/Polygon';
+
+
```
LukasLohoff, line 16:
```suggestion
```

src/Button/DrawButton/DrawButton.tsx
```diff
@@ -0,0 +1,308 @@
...
+        setShowLabelPrompt(true);
+        evt.feature.set('isLabel', true);
+        setDigitizeTextFeature(evt.feature);
+      });
+    }
+
+    return () => {
+      unByKey(key);
+      map.removeInteraction(interaction);
+    }
```
LukasLohoff, line 200:
```suggestion
    };
```

LukasLohoff (approved):
:medal_sports:

src/Button/DrawButton/DrawButton.tsx
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * Currently drawn feature which should be represent as label or postit.
+   */
+  const [digitizeTextFeature, setDigitizeTextFeature] = useState<OlFeature<OlGeometry>>(null);
+
+  /**
+   * Called when the digitize button is toggled. If the button state is pressed,
+   * the appropriate draw, modify or select interaction will be created.
+   * Otherwise, by untoggling, the same previously created interaction will be
+   * removed from the map.
```
annarieger, line 224: I think this comment is a bit misleading

src/Button/DrawButton/DrawButton.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+  return (
+    <span className={btnWrapperClass}>
+      <ToggleButton
+        onToggle={onToggleInternal}
+        className={finalClassName}
+        {...passThroughProps}
+      />
+      {
+        showLabelPrompt ?
```
annarieger, line 278: absolutely minor:

can be simplified to `showLabelPrompt && <Modal [...] />`

src/Util/DigitizeUtil.ts
```diff
@@ -0,0 +1,220 @@
...
+      }),
+      stroke: new OlStyleStroke({
+        color: DigitizeUtil.DEFAULT_STROKE_COLOR
+      })
+    })
+  });
+
+  static DIGITIZE_LAYER_NAME = 'react-geo_digitize';
+
+  static getDigitizeLayer(map: OlMap): OlVectorLayer<OlSourceVector<Geometry>> {
```
annarieger, line 86: A tiny comment would be useful :)

annarieger (approved):
Very nice work! 🍍

PR #2316, opened 2021-08-27
Migrate ZoomButton and AddWmsLayerEntry tests

LukasLohoff:
### Description:

Migrates ZoomButton and AddWmsLayerEntry tests to work with testing-library.

@terrestris/devs Please review

### DIFF
--- a/src/Button/ZoomButton/ZoomButton.spec.tsx
+++ b/src/Button/ZoomButton/ZoomButton.spec.tsx
@@ -1,6 +1,10 @@
 import TestUtil from '../../Util/TestUtil';
 import OlMap from 'ol/Map';
 
+import { render, screen } from '@testing-library/react';
+import * as React from 'react';
+import userEvent from '@testing-library/user-event';
+
 import ZoomButton from './ZoomButton';
 
 describe('<ZoomButton />', () => {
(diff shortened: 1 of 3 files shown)


src/Button/ZoomButton/ZoomButton.spec.tsx
```diff
@@ -16,60 +20,71 @@ describe('<ZoomButton />', () => {
...
+      <ZoomButton map={map} delta={-1}>Zoom test</ZoomButton>
+    );
 
     const initialZoom = map.getView().getZoom();
-
-    wrapper.instance().onClick();
+    const button = screen.getByText('Zoom test');
+    userEvent.click(button);
 
     const promise = new Promise(resolve => {
```
dnlkoch, line 55: Not related to the changes in particular, but we may want to make use of ours [actSetTimeout()](https://github.com/terrestris/react-geo/blob/master/src/Util/rtlTestUtils.tsx#L8) here.

dnlkoch (approved):

src/Button/ZoomButton/ZoomButton.spec.tsx
LukasLohoff: didn't know that existed, fixed

PR #2317, opened 2021-08-30
Add SelectFeaturesButton

simonseyock:
### Description:
This adds a button that gets passed an array of layers from which to select vector features from. This is a building block for the upcoming CopyButton, DeleteButton and ModifyButton.

It renames the `renderInContext` method to `renderInMapContext` to make it a little clearer.

### DIFF
--- a/src/Button/SelectFeaturesButton/SelectFeaturesButton.tsx
+++ b/src/Button/SelectFeaturesButton/SelectFeaturesButton.tsx
@@ -1,16 +1,18 @@
+import * as React from 'react';
+import { useEffect, useRef } from 'react';
+
 import OlInteractionSelect, { Options as OlSelectOptions, SelectEvent as OlSelectEvent } from 'ol/interaction/Select';
 import { StyleLike as OlStyleLike } from 'ol/style/Style';
-import VectorLayer from 'ol/layer/Vector';
-import VectorSource from 'ol/source/Vector';
-import Geometry from 'ol/geom/Geometry';
-import ToggleButton, { ToggleButtonProps } from '../ToggleButton/ToggleButton';
-import { useEffect, useState } from 'react';
-import { useMap } from '../..';
+import OlVectorLayer from 'ol/layer/Vector';
+import OlVectorSource from 'ol/source/Vector';
+import OlGeometry from 'ol/geom/Geometry';
 import * as OlEventConditions from 'ol/events/condition';
-import { DigitizeUtil } from '../../Util/DigitizeUtil';
 import { unByKey } from 'ol/Observable';
+
+import ToggleButton, { ToggleButtonProps } from '../ToggleButton/ToggleButton';
+import { useMap } from '../../Hook/useMap';
+import { DigitizeUtil } from '../../Util/DigitizeUtil';
 import { CSS_PREFIX } from '../../constants';
-import * as React from 'react';
 
 interface OwnProps {
   /**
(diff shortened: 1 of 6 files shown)


src/Util/rtlTestUtils.tsx
```diff
@@ -48,24 +59,33 @@ export function clickMap(map: OlMap, x: number, y: number) {
...
+
+  const { rerender, ...results } = render(assemble(element));
 
   map.setSize([400, 400]);
   map.renderSync();
 
-  return result;
+  const rerenderInMapContext = (element: ReactElement) => {
+    rerender(assemble(element));
+  }
```
dnlkoch, line 85:
```suggestion
  };
```

src/Util/rtlTestUtils.tsx
```diff
@@ -48,24 +59,33 @@ export function clickMap(map: OlMap, x: number, y: number) {
...
-    <MapContext.Provider value={map}>
+export function renderInMapContext(map: OlMap, element: ReactElement, size: [number, number] = [400, 400]) {
+  const assemble = (element: ReactElement) => {
+    return <MapContext.Provider value={map}>
       <MapComponent map={map} />
       {element}
-    </MapContext.Provider>
-  );
+    </MapContext.Provider>;
+  }
```
dnlkoch, line 76:
```suggestion
  };
```

src/Util/rtlTestUtils.tsx
```diff
@@ -48,24 +59,33 @@ export function clickMap(map: OlMap, x: number, y: number) {
...
 
 /**
  * This function renders the given element inside a map context and initializes the map with size `[400, 400]`, ready
  * to be used by the event functions in this file.
  */
-export function renderInContext(map: OlMap, element: ReactElement) {
-  const result = render(
-    <MapContext.Provider value={map}>
+export function renderInMapContext(map: OlMap, element: ReactElement, size: [number, number] = [400, 400]) {
+  const assemble = (element: ReactElement) => {
```
dnlkoch, line 71: `element` is already declared in the upper scope.

src/Util/rtlTestUtils.tsx
```diff
@@ -48,24 +59,33 @@ export function clickMap(map: OlMap, x: number, y: number) {
...
+    </MapContext.Provider>;
+  }
+
+  const { rerender, ...results } = render(assemble(element));
 
   map.setSize([400, 400]);
   map.renderSync();
 
-  return result;
+  const rerenderInMapContext = (element: ReactElement) => {
```
dnlkoch, line 83: `element` is already declared in the upper scope.

src/Button/SelectFeaturesButton/SelectFeaturesButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+      layers: layers,
+      ...(selectInteractionConfig ?? {})
+    });
+
+    newInteraction.set('name', selectInteractionName);
+    newInteraction.setActive(false);
+
+    map.addInteraction(newInteraction);
+
+    setSelectInteraction(newInteraction);
```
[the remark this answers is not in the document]
simonseyock: A module variable is not an option because then every DrawButton would use the same interaction which will lead to unwanted sideeffects.
I was not thinking about `useRef` while implementing. In the current implementation the binding of the eventlistener is done in another `useEffect` do the interaction is used as dependency there. But I think it is not a big issue to recreate the interaction if the listener changes. This is done for many prop changes anyways (for example `style`, `interactionConfig`, ...).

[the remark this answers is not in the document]
simonseyock: I put everything in the same `useEffect` and used `useRef` now.

dnlkoch (approved):

Issue #2327, opened 2021-12-13
Add geodesic option on measure button tool


jkeruzec:
Is it possible to add geodesic option on measure tool to have this option available? 

Maybe you want a PR for this ? 

Regards,

dnlkoch:
Available via [Release 15.4.0](https://github.com/terrestris/react-geo/releases/tag/v15.4.0).

PR #2328, opened 2021-12-13
#2327 add geodesic option for measure tool

jkeruzec:


jkeruzec:
@dnlkoch How I can go further with the PR ? My comment isn't good enough :(

dnlkoch:
Hi @jkeruzec, thanks for your very welcome contribution! 

We make use of [semantic pull requests/commits](https://github.com/zeke/semantic-pull-requests) in react-geo which requires that all commits should be prefixed with the semver type. In your case I suggest to prefix the commit with `feat`:

```
terrestris#2327 add geodesic option for measure tool
```
⤋
```
feat: add geodesic option for measure tool
```

Your proposed diff looks quite nice, so thanks again!

PR #2688, opened 2022-07-19
feat: allow ordering searchResult categories

jansule:
## Description

This refactors `<SearchResultsPanel>` to allow specifying the order of the result categories. At the same time, the property `features` was renamed to `searchResults` for better readability.

BREAKING CHANGE: The property `features` was replaced with
`searchResults` and its type was changed to `Category[]`.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist

- [ ] I have added or updated tests and documentation, and the test suite passes (run `npm test` locally).

### DIFF
--- a/src/Panel/SearchResultsPanel/SearchResultsPanel.tsx
+++ b/src/Panel/SearchResultsPanel/SearchResultsPanel.tsx
@@ -18,10 +18,13 @@ import BaseLayer from 'ol/layer/Base';
 const Panel = Collapse.Panel;
 const ListItem = List.Item;
 
+interface Category {
+  title: string;
+  features: OlFeature[];
+}
+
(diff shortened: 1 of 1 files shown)


dnlkoch (approved):

KaiVolland:
You could avoid a breaking change if you would leave the `features` prop and may just deprecate it.

jansule:
> You could avoid a breaking change if you would leave the `features` prop and may just deprecate it.

I did not only change the property name, but also its structure/type. So there would be a breaking change anyway. The same functionality could also have been handled with an additional property, but since this component is only 19 days old, I went for the cleaner approach, code-wise

PR #2921, opened 2022-10-25
test: add some tests for isWmsLayer

marcjansen:
## Description

Originally I just wanted to increase coverage, and picked sth. supposedly trivial to test, but then I stumbled about the logic behind `isWmsLayer`:

- Is this needed in a typed world at all?
- if it is needed, is it likely used in many projects?
- does it make sense the way it is now, so that e.g. a vector layer with a wms source is considered a wms layer? The longer I look at the code the more wrong it looks to my eye.

## Related issues or pull requests

-

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [x] Other (please describe): tests

- [ ] Yes
- [x] No, right now not, but it might be we remove the function or change it's behaviour?

## Checklist

### DIFF
--- a/src/Util/typeUtils.spec.tsx
+++ b/src/Util/typeUtils.spec.tsx
@@ -1,0 +1,130 @@
+import { isWmsLayer } from './typeUtils';
+
+import OlBaseLayer from 'ol/layer/Base';
+import OlLayer from 'ol/layer/Layer';
+import OlImageWMS from 'ol/source/ImageWMS';
+import OlTileWMS from 'ol/source/TileWMS';
+import OlImageLayer from 'ol/layer/Image';
+import OlTileLayer from 'ol/layer/Tile';
+
+import OlVectorLayer from 'ol/layer/Vector';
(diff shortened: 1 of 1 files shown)


KaiVolland:
IMHO this util – and the new and nice tests – should be moved to the [`ol-util`](https://github.com/terrestris/ol-util) as ther is no connection to react.

marcjansen:
Sobyou think it still makes sense at all? And is ol-util already written in ts?

KaiVolland:
Maybe we can merge this one here and create an issue to move stuff without react connection to the related packages? :thinking:

marcjansen:
Works for me, especially, since the method is not at all used internally in this library.

marcjansen:
The mentioned issue is here: https://github.com/terrestris/react-geo/issues/2923 We can first gather candidates and then move things one by one. Should we first merge this? If so, I'd favor a :+1: and a review ;-)

marcjansen:
Thanks let's do it like described.

Issue #2298, opened 2021-07-02
Use react testing library instead of enzyme


simonseyock:
The react testing library enables to test components a lot like users would use them. Also it does not need full applications, it can just render single components and test them as if they were in a real page.

It is still quite atomic and can therefore replace the common tests formerly written in enzyme. It follows another philosophy though, that I find quite reasonable (see below). It enables to work with newer features of redux, for example the `useSelector` hook.

This text is from the react testing library page about enzyme:

> The primary purpose of the React Testing Library is to give you confidence by testing your components in the way the user will use them. Users don't care what happens behind the scenes. They just see and interact with the output. So, instead of accessing the components' internal API, or evaluating the state, you'll get more confidence by writing your tests based on the component output.
> 
> React Testing Library aims to solve the problem that many developers face when writing tests with Enzyme which allows (and encourages) developers to test implementation details. Tests which do this ultimately prevent you from modifying and refactoring the component without changing the test. As a result, the tests slowed down your development speed and productivity, since every small change requires rewriting some part of your tests, even if the change you made does not affect the component's output.
> 
> Re-writing your tests in React Testing library is worthwhile, because you're trading tests that slow you down for tests that give you more confidence and increase your productivity in the long run.

https://testing-library.com/docs/react-testing-library/migrate-from-enzym

### Find tests

1. What user actions need to be done to test the functionality that was formerly tested with enzyme?
2. Look in the styleguide, is there e2e functionality that could be tested additionally?
3. Look at the props of the component, is there something that should be tested as well?

### Write tests

1. Use `getRole` / `getText` / etc. to retrieve elements. Try to be as semantic as possible. Look at the errors of these functions. They help a lot. If needed add `role` tags, this helps accessibility of the page as well (see point 5 for antd problems). Do only use `query...` methods if you want to test if an element is not available.
2. Use `@testing-library/jest-dom` (i.e. `expect(input).toHaveValue('some value');` or `expect(element).toBeVisible();`) for better test output.
3. Use `jest-fetch-mock` to mock fetch
4. Use `waitFor` or `findRole` / `findText` / etc. to work with async effects. They help with `not wrapped in act(...)` and `performed state update on unmounted component` warnings. These should all be resolved and can sometimes even occur only if multiple tests are run after each other. For special cases there exists `TestUtil.setTimeout`, try to avoid it.
[...]

PR #3074, opened 2023-01-10
feat(NominatimSearch): rewrite to FC, add debounceTime, use rtl for test

simonseyock:
## Description

* Rewrite of `NominatimSearch` to be a Functional Component
* Adds a `debounceTime` prop (default: 300 ms) so that no requests are made while the user is still typing
* rewrites the test to use react-testing-library
* rewrites the example to use a map context and a FC

## Related issues or pull requests

https://github.com/terrestris/react-geo/issues/2298

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/Field/NominatimSearch/NominatimSearch.example.md
+++ b/src/Field/NominatimSearch/NominatimSearch.example.md
@@ -1,25 +1,22 @@
 This demonstrates the usage of the NominatimSearch.
 
 ```jsx
-import * as React from 'react';
+import {useEffect, useState} from 'react';
 
 import OlMap from 'ol/Map';
 import OlView from 'ol/View';
 import OlLayerTile from 'ol/layer/Tile';
 import OlSourceOSM from 'ol/source/OSM';
-import { fromLonLat } from 'ol/proj';
+import {fromLonLat} from 'ol/proj';
 
+import MapComponent from '@terrestris/react-geo/Map/MapComponent/MapComponent';
 import NominatimSearch from '@terrestris/react-geo/Field/NominatimSearch/NominatimSearch';
 
-class NominatimSearchExample extends React.Component {
+const NominatimSearchExample = () => {
+  const [map, setMap] = useState();
 
-  constructor(props) {
-
-    super(props);
-
-    this.mapDivId = `map-${Math.random()}`;
-
-    this.map = new OlMap({
+  useEffect(() => {
+    const newMap = new OlMap({
       layers: [
         new OlLayerTile({
           name: 'OSM',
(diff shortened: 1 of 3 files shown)


KaiVolland:
Cool stuff but I'm not sure about using the `useMap` hook inside the react-geo components. This forces you to use the `mapContext` and you loose some flexibility.

simonseyock:
In many other components it is used this way. We should decide on a consistent way to do this. Maybe adapt `useMap` that it accepts an optional `map` param that is returned if no context is detected? Than we could pass in the map param from the component and make the param optional and use this in all components.

KaiVolland:
> In many other components it is used this way. We should decide on a consistent way to do this. Maybe adapt `useMap` that it accepts an optional `map` param that is returned if no context is detected? Than we could pass in the map param from the component and make the param optional and use this in all components.

IMHO this would be even more confusing. Today I'd prefer to not use the `useMap` hook inside of the components. But your totaly right that we should have a consistent handling for this over all components.

simonseyock:
ok, i thought about something like this:

```
export const useMap = (map?: OlMap): OlMap|null => {
  const contextMap = useContext(MapContext);
  return map ?? contextMap;
};

const Comp = ({
  map
}) => {
  const actualMap = useMap(map);
  // ...
};
```

simonseyock:
Ok, then we could remove the `MapContext` and `useMap` completely because it is only trivial code that provides not much benefit.

dnlkoch:
Why do you want to remove it? Many applications depend on it.

simonseyock:
It is only a possibility to remove it, because it is then totally independent from `react-geo` and are only like 10 lines of code.

simonseyock:
If it was removed it would be very clear that it is not needed by `react-geo `components and that it can be removed if it is not used in the application itself.

simonseyock:
I readded the map property and removed the `BREAKING CHANGE` commit body.

dnlkoch:
In terms of backwards compatibility I wouldn't recommend to remove the hook. It's part of the stable API - even if not used internally - for a long time and is used in several applications that would need to copy it over. I don't see any benefit of removing it.

simonseyock:
ok. so the tendency is to remove all internal use of `useMap`, but still provide it for applications which might need it.

simonseyock:
But in the meantime, can this PR get merged?

dnlkoch (approved):

Issue #2544, opened 2022-05-11
[Bug]: Incorrect import Paths in Documentation


jansule:
### Current Behavior

Import paths in the [react-geo documentation](https://terrestris.github.io/react-geo/) seem to be wrong. The `/dist/` part is missing.

Examples (copied from https://terrestris.github.io/react-geo/docs/latest/#/Components/Buttons/CopyButton):

```
import MapContext from '@terrestris/react-geo/Context/MapContext/MapContext'
import MapComponent from '@terrestris/react-geo/Map/MapComponent/MapComponent';
import CopyButton from '@terrestris/react-geo/Button/CopyButton/CopyButton';
import {DigitizeUtil} from '@terrestris/react-geo/Util/DigitizeUtil';
```

### Expected Behavior

Import paths should contain the `/dist/` part right after `@terrestris/react-geo/`, i.e. 

```
import MapContext from '@terrestris/react-geo/dist/Context/MapContext/MapContext'
import MapComponent from '@terrestris/react-geo/dist/Map/MapComponent/MapComponent';
import CopyButton from '@terrestris/react-geo/dist/Button/CopyButton/CopyButton';
import {DigitizeUtil} from '@terrestris/react-geo/dist/Util/DigitizeUtil';
```

### Steps To Reproduce

1. Include react-geo as dependency into your project
2. Copy any example from the docs
3. See it not working

### Version

16.2.0

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3149, opened 2023-02-17
fix: example imports

weskamm:
This closes https://github.com/terrestris/react-geo/issues/2544 by rewriting the import pathes to correctly use the `dist` folder.

In order to keep styleguidist serving those examples live with hot reloading while developing, the alias has been adjusted to load classes from the src instead of this dist folder.


annarieger (approved):
Thanks! 👍🏻

PR #3715, opened 2024-01-10
Configurable onClick property for SearchResultPanel

mholthausen:
## Description

This PR adjusts the `SearchResultPanel` with an optional property (`extentShift`) to apply  an offset to the features extent of an entry in the search results. This can be used when the map is overlaid by elements and the map center is no longer the actual map center at that moment.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/Panel/SearchResultsPanel/SearchResultsPanel.tsx
+++ b/src/Panel/SearchResultsPanel/SearchResultsPanel.tsx
@@ -36,7 +36,7 @@ interface SearchResultsPanelProps extends Partial<CollapseProps> {
   /** A renderer function returning a prefix component for each list item */
   listPrefixRenderer?: (item: any) => undefined | JSX.Element;
   layerStyle?: undefined | OlStyle;
-  extentShift?: number | undefined;
+  onClick?: (item: any) => void;
 }
 
 const SearchResultsPanel = (props: SearchResultsPanelProps) => {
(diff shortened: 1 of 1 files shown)


KaiVolland:
Wouldn't it be better to add an `onClick` property which zooms to the feature by default and can then be overwritten? This looks like a quite specific requirement.

mholthausen:
Yes its specific requirement that was used in another project before and should be ported to the `SearchResultPanel` now so that it can be used by the gis client.

KaiVolland (changes requested):
I added some suggestions to make clearer what i meant. (Code is not tested.)

This way you keep the project specific logic outside of this library *and* you make it even more flexible for other projects.

dnlkoch (changes requested):
I agree with @KaiVolland here. Please make the `onClick` prop configurable to it can be adjusted to whatever an application might need.

mholthausen:
Okay, added a configurable `onClick` property in this component in [62a40c4](https://github.com/terrestris/react-geo/pull/3715/commits/62a40c469f7816be5b8af8b261e29da24ae6636c) with the previous behavior which can be overwritten.

KaiVolland (approved):
nice. thanks for the adjustments.

mholthausen:
:+1: thanks for the reviews!

PR #3931, opened 2024-07-31
feat: adds a column filter

FritzHoing:
## Description

This MR adds a column filter that can also be used to define the column order. If the attribute is set, all columns defined in the array will be displayed. If the attribute is not set, all columns will be shown.

Attributes defined in the blacklist will be ignored from the list anyways.

Dragging a column works as expected.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [ ] I have added or updated tests and documentation, and the test suite passes (run `npm test` locally).

### DIFF
--- a/src/Grid/FeatureGrid/FeatureGrid.tsx
+++ b/src/Grid/FeatureGrid/FeatureGrid.tsx
@@ -51,6 +51,10 @@ type OwnProps = {
    * When active the order of the columns can be changed dynamically using drag & drop.
    */
   draggableColumns?: boolean;
+  /**
+   * Array of dataIndex names to filter
+   */
+  attributeFilter?: string[];
   onRowSelectionChange?: (selectedRowKeys: Array<number | string | bigint>,
     selectedFeatures: OlFeature<OlGeometry>[]) => void;
 };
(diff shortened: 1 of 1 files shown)


src/Grid/FeatureGrid/FeatureGrid.tsx
```diff
@@ -294,18 +299,28 @@ export const FeatureGrid = <T extends AnyObject = AnyObject,>({
...
     const columnDefs: ColumnsType<T> = [];
-    if (features && features.length < 1) {
+    if (!features || features.length < 1) {
       return;
     }
 
-    const feature = features?.[0];
-    const props = feature?.getProperties();
+    const feature = features[0];
+    let props = feature?.getProperties();
```
simonseyock, line 307:
```suggestion
    const props = feature?.getProperties();
```

FritzHoing:
> Very nice!
> 
> I was thinking if it would be nicer to also name the property `attributeFilter` instead of `columnFilter`, because it is technically one, the PropertyGrid has the same attribute and the blacklist is also prefixed with `attribute`
> 
> Would be great if you could add the `attributeFilter` to https://github.com/terrestris/react-geo/blob/main/src/Grid/commonGrid.ts and implement it for AgFeatureGrid also

Thank you for the review!

Yes i will do that in a follow up.

simonseyock (approved):

PR #4202, opened 2025-03-28
Fixes a zoom error in LayerSwitcher

mholthausen:
## Description

This PR fixes a bug in the zoom behavior of the main map for v24 when the `LayerSwitcher` is included. Previously a too small scale was reached when zooming in. The user expected to zoom closer to the selected area by dragging a box.  
This is achieved by synchronizing the view between the main map and the preview map, instead of using the same view:

Old behavior:  
old

New behavior:  
new

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [x] I have added or updated tests and documentation, and the test suite passes (run `npm run check` locally).
- [x] I have added a screenshot/screencast to illustrate the visual output of my update.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,7 @@ on:
       - main
       - next
       - 15.x.x
+      - 24.x.x
 
 jobs:
   release:
--- a/.releaserc
+++ b/.releaserc
@@ -1,6 +1,7 @@
 {
   "branches": [
     "main",
+    "24.x.x",
     {
       "name": "next",
       "prerelease": true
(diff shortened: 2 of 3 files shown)


weskamm:
I am confused. Why does the previewmap seem to display something completely different than the main map?
I expected the previewmap (overview?) to show the same center with a smaller scale to get a better overview, but it looks somehow totally different?

mholthausen:
@weskamm I was also a bit confused when I started working with the component, but the preview map in the LayerSwitcher always shows the next possible background map and not the currently selected one. This is actually intentional.

weskamm:
Yes thats fine, but it seems the center and zoom of it seems to be completely off

weskamm:
I am not exactly sure why this change fixes the issue, but it may be a good idea anyways to have a map-view clone instead of using the same instance to avoid other bugs or "accidents".
I think the underlying problem is related to fixed resolutions and a constrain that forces openlayers to not zoom to the exact bbox but the next best matching instead. It may be related to using the ol/source/osm, but there may also be other reasons.
Anyways, if this fixes the issue, i will approve!

weskamm (approved):

